PR #1, opened 2019-02-12
Travis Setup

ysimillides:
Currently the build fails ( https://travis-ci.org/ysimillides/MLJModels.jl ) but nevertheless, it would be a good idea to get this setup so we can check this moving forward.


ablaom:
Travis was set up and builds fine.

PR #2, opened 2019-02-13
[WIP] GLM

tlienart:
### Implementation

* [x] add generic GLM for arbitrary Distribution/Link
* [x] add basic tests
* [ ] GLM constructor should also take parameters with max number of iterations etc
* [ ] check metadata is fine (@ablaom)

### Clashing with rest of repo

* [ ] `GLMRegressor.distribution` should be a `Distribution` type but that requires modifying the Manifest to have `Distributions` in the requirements (@ablaom)

**Notes**: 
- working branch is wip-glm but due to the hacks in `runtests.jl` and `*.toml` files I'm using this other branch to only port the changes to `src/GLM.jl` and `test/GLM.jl`.
- this passes tests locally but might not on Travis since we're not using the same `*.toml` files.
- if there's a step-by-step process to modify `*.toml` files without it clashing, I'll happily follow it.

### DIFF
--- a/src/GLM.jl
+++ b/src/GLM.jl
@@ -2,47 +2,57 @@ module GLM_
 
 import MLJBase
 
-export OLSRegressor, OLS, LinearRegression
+export OLSRegressor, OLS,
+       GLMRegressor, GLM
 
 import ..GLM # strange syntax for lazy-loading
+import ..Distributions
 
-const OLSFitResult = GLM.LinearModel
+const LMFitResult  = GLM.LinearModel
+const GLMFitResult = GLM.GeneralizedLinearModel
 
-OLSFitResult(coefs::Vector, b=nothing) = OLSFitResult(coefs, b)
+LMFitResult(coefs::Vector, b=nothing) = LMFitResult(coefs, b)
+GLMFitResult(coefs::Vector, b=nothing) = GLMFitResult(coefs, b)
 
-####
-#### OLSRegressor
-####
-
-mutable struct OLSRegressor <: MLJBase.Deterministic{OLSFitResult}
+mutable struct OLSRegressor <: MLJBase.Deterministic{LMFitResult}
     fit_intercept::Bool
 #    allowrankdeficient::Bool
 end
 
-function OLSRegressor(;fit_intercept=true)
-    return OLSRegressor(fit_intercept)
+OLSRegressor(;fit_intercept=true) = OLSRegressor(fit_intercept)
+
+# shorthand
+const OLS = OLSRegressor
+
+mutable struct GLMRegressor <: MLJBase.Deterministic{GLMFitResult}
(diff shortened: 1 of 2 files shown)


ablaom:
Many thanks for that. 

Not sure what the issue with distributions is that you mention. Distributions is a dependency ofGLM which re-exports the distributions it needs, so we don't need to explicitly import it. 

If you need to import packages for testing (eg, Distances is needed for test/Clustering.jl) then just `add` them to the Project.toml using the package manager. 

I see that MLJ got added back to the Project.toml file, which means MLJ cannot get installed in Julia because of a circular dependency - at least that was my experience. But the `Pkg.add(...)` in runtests.jl means you should be able to remove MLJ from MLJModels/Project.toml, as on master. Remember to run `resolve` after any manual changes to Project.toml.

Everything looks good so far with one key comment: The idea is that predictions should be **probabilistic** for this case, no? Your `predict` should predict one distribution for each input pattern, ie a vector of distributions. The distribution type matches the hyperparameter value of `distribution`. And your model subtype is then `Probabalistic{LMFitResult}` instead of `Deterministic{LMFitResult}`. (If point-predictions are required, the user gets them with a fall-back `predict_mean` or `predict_median` or adds his own post-predict transformer. See [issue #34](https://github.com/alan-turing-institute/MLJ.jl/issues/34) and MLJ/doc/adding_new_models.md).

I agreed with @fkiraly (see #34) to have separate models for each "output_kind" (see the list in the adding_new_models.md guide). So, technically, Poisson (for input_kind = :ordered_factor_infinite) should get its own model, while Normal, Gamma, and so forth (for input_kind = :continuous) can have their own model. Perhaps call the first one GLMPoisson (or GLMCount??) and the other GLMRegressor. Handling the other output_kinds :mulitclass, :binary and :ordered_factor_finite should wait for a discussion that I will start on MLJ soon.

Another minor comment is that your LMFitResult type is not a concrete type. I think this will more-or-less kill the performance gains I hope for in the ensembling. At some point we you might want to figure out what the concrete type is and you may have to add type parameters to your model (eg, input and/or target element type) to do this. This is admittedly a pain, and might be better left as a later optimisation.

tlienart:
* in the `GLMRegressor`, I'd like `distribution::Distributions.Distribution` instead of `distribution` (`::Any` type), this requires importing `Distribution`. I'm reluctant to bring my modifications of the `project.toml` and `manifest.toml` to this PR as I believe the current situation to be brittle (as per the `Pkg.add` line  in `runtests` which shouldn't be there; I've removed it locally). I don't think this needs to be solved now, when the package dependencies have been figured out a bit more, we can go back to it.
* Probabilistic output, I'll look into it.

I guess it's probably easier if we close this PR then and I suggest another one with just Poisson link.

PR #3, opened 2019-02-14
GLMCount + probabilistic output

tlienart:
Follow up from #2 cleaned up so that

**key**

* `predict(::OLS)` outputs a vector  of Normal
* `predict(::GLMCount)` outputs a  vector of Poisson
* `predict_mean` works as pointwise predict

**minor**

* I included changes to `*.toml` files, hopefully Travis will be ok with it (looks like it is, cool!)
* metadata fixed as per comments in #2 
* waiting for a pending discussion on binary outputs etc to implement `GLMBinary` etc (with Binomial/Probit and all that jazz)

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -19,41 +19,17 @@ git-tree-sha1 = "055eb2690182ebc31087859c3dd8598371d3ef9e"
 uuid = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
 version = "0.5.3"
 
-[[CSV]]
-deps = ["CategoricalArrays", "DataFrames", "DataStreams", "Dates", "Mmap", "Parsers", "Profile", "Random", "Tables", "Test", "Unicode", "WeakRefStrings"]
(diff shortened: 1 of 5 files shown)


ablaom:
Great but please fix your Project file conflicts. Since GLM is already in the right place ( in [extras]) all you should need to do is `]add Distributions` to the Project file on master, no?

ablaom:
Also, do not use model aliases in `load_path`s. These must refer to the literal struct name for MLJ's @load to work properly. Thanks

ablaom:
https://github.com/alan-turing-institute/MLJ.jl/issues/81#issuecomment-465834053

tlienart:
Ok I think that's good to go. I kept the models separated as I think refactoring for something like `GLM{continuous}` etc should not be very hard and is probably best left for when we have a clearer vision on what is desired

ablaom:
Great, thanks!

ablaom:
MLJRegsitry is now updated. You should be able to:

````julia
using MLJ
@load GLMCountRegressor
````

PR #6, opened 2019-02-25
Decision tree

ysimillides:
The tests where taken from the DecisionTree test suite, they mightneed to be changed to fit our specification

### DIFF
--- a/src/DecisionTree.jl
+++ b/src/DecisionTree.jl
@@ -21,7 +21,7 @@ using CategoricalArrays
 #> import package:
 import ..DecisionTree # strange syntax b/s we are lazy-loading
 
-# here T is target type, and the `Vector{T}` is for storing target levels:  
+# here T is target type, and the `Vector{T}` is for storing target levels:
 const DecisionTreeClassifierFitResultType{T} =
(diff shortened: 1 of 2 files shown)


ablaom:
Looking great, thanks!

- What you have implemented is Deterministic, not Probabilistic, ie you are predicting point-values not distributions. That's fine, but you should change `<: Probabilistic{Any}` to `<: Deterministic{Any}` (I'll fix `Any` later). Note: the classifier *is* presently probabilistic, but this currently can't be made to work in the regression case (for reasons to do with not being able to fit most continuous distributions to a single data point).

- There's a typo in src/DecisionTree.jl, line 251:
`MLJBase.load_path(::Type{<:DecisionTreeRegressor}) = "MLJModels.DecisionTree_.DecisionTreeClassifier"`

- currently the tests provides `features` as a matrix. It should be a table. So you could do something like `features = MLJBase.table(features)`.

- to test pruning, please add a test with baretree.post_prune = true`. 

- add `info(R1tree)` to test metadata

- add `@test model1 isa MLJBase.fitresult_type(R1Tree)` to test fit result type is correct.

PR #10, opened 2019-03-26
XGBoost

ysimillides:


ablaom:
I'm still having problem with negative `silent`. I should have been clearer, sorry: Instead of

```
    if(verbosity>1)
        verbosity==1
```
use

```
verbosity = 
    verbosity > 0 ?  1 : 0

```

Also, you can remove `silent` as field of all model structs; it is, as you suggest, redundant. 

Your HDF5 build error looks kind of random. Not sure what that is. 

Thanks

ablaom:
Thanks for that. 

Comments:

- on line 60 of test/XGBoost.jl `clpred` is defined. Tests pass but the value I get is weird (see below). It is supposed to be a vector of distributions not a categorical vector (and it should have sensible values). I recall you working on code to define the distributions but can't find it in the current commit??

```
julia> clpred
180-element CategoricalArray{String,1,UInt32}:
 #undef
 #undef
 #undef
 #undef
 ...
```

- line 61 is not an appropriate test. replace with 

```
@test levels(clpred[1]) == levels(y[train])

```

- we did not discuss this, but XGBoost probably treats the two-class (binary) class differently. So while 3 levels (classes) means 3 probabilities per input row in predict, 4 levels means 4 probabilities per row, and so forth, I expect (but you should check) that for 2 levels only 1 probability is output (and the second is 1 minus this one). If so, you have a case distinction. To test the binary case, repeat the same deterministic classifier test you have but with iris data replaced with some synthetic data, say:

```
X = (x1=rand(100), x2=rand(100))
y = map(X.x1) do x mod(10*x, 2) end
```

ablaom:
Finished implementing above comments and merged.

PR #15, opened 2019-05-10
implement elastic net alongside tests.

ysimillides:


ablaom:
That' great. Few details:

- Please use the names given in the issue, eg, `ElasticNet` not `SCElasticNet`. (Note duplicate model names are fine if the models come from different packages.)

-  `dual_gap` is missing from the report for `ElasticNetCV`

` the report keys for `ElasticNet` should be  `(:n_iter, :dual_gap)`. I've correct my ambiguity at

ysimillides:
if you keep the names the same, you get a naming clash with 
@sk_import linear_model : ElasticNet 
@sk_import linear_model : ElasticNetCV
and then
LoadError: invalid redefinition of constant ElasticNet.
I haven't figured out how to avoid this, but if you have any ideas I can change them from SCElastic to Elastic (Although I admit I might just be doing something completely wrong with the import paths.)

ablaom:
Thanks for that. For future reference, the solution to the name conflicts is to substitute the `@sk_import` calls with this code:

```
(ScikitLearn.Skcore).import_sklearn()    # just needed once in the ScikitLearn_ module
ElasticNet_ =
    (ScikitLearn.Skcore.pyimport)("sklearn.linear_model")[:ElasticNet]
ElasticNetCV_ =
    (ScikitLearn.Skcore.pyimport)("sklearn.linear_model")[:ElasticNetCV]
 ```

The underscore versions are what you then use to construct the things you called "cache".

PR #31, opened 2019-08-13
Update of the GLM models, OLS+Binary

tlienart:
* Prob OLS regression
* Prob Binary classification (probit, logit, ...)

### Future 

(will be another pr)

* add count regression (poisson)
* add multinomial classif (will be for another PR, after https://github.com/alan-turing-institute/MLJ.jl/issues/175) 

### Notes

* decode is now used properly, thanks for the feedback
* MLJBase 0.4 seems to require different metadata, (`scitype_union` -> `scitype` and no `input_multivariate` etc) so I added a few fixes for those; given that 0.4 has not been merged but will soon, I've added a statement that checks the version and acts accordingly; it'll be easy to remove later on.
  * this incurred a few missed commits where things were passing locally but not on Travis, I'll bring this up in the next call, it mostly seems to be a versioning issue.


tlienart:
I'll wait until you've pushed your changes corresponding to 0.4; then will merge master here, cleanup and re-submit :)

tlienart:
Ok it passes on Julia 1 and 1.1 on Travis. It fails on nightly apparently due to an issue with conda installation as far as I can tell. So it should be good for reviewing @ablaom

src/GLM.jl
```diff
@@ -1,119 +1,151 @@
...
-
-    Xmatrix = MLJBase.matrix(X)
-    features = MLJBase.schema(X).names
-    model.fit_intercept && (Xmatrix = hcat(Xmatrix, ones(eltype(Xmatrix), size(Xmatrix, 1), 1)))
-
-    fitresult = GLM.lm(Xmatrix, y)
+# NormalRegressor       --> Probabilistic w Continuous Target
+# PoissonRegressor      --> Probabilistic w Count Target
+# BernoulliClassifier   --> Probabilistic w Binary target // logit,cauchit,..
+# MultinomialClassifier --> Probabilistic w Multiclass target
```
ablaom: I guess you mean "BinaryClassifier" not "BernoulliClassifier" here?

src/GLM.jl
```diff
@@ -1,119 +1,151 @@
...
+import Distributions
+using Parameters
 
 import ..GLM
 
-####
-#### REGRESSION TYPES
-####
+export NormalRegressor, OLSRegressor, OLS,
+        BinaryClassifier
```
ablaom: You export four models but you define only two structs. Currently the ordinary user will not see these aliases. Also, the proliferation of aliases makes it confusing to review the code : -)

tlienart:
Ok, I've implemented your suggestions; will wait for tests to pass and merge.

Re Count, I'll do that in another PR;

Re Multiclass, is that supported now? Could you point me to the equivalent of `MultivariateFinite`? thanks!

ablaom:
Nevermind, about MutlivariateFinite; I had forgot our earlier discussion. 

I'm puzzled why you have moved LIBSVM from optional dependency to actual dependency.

PR #26, opened 2019-07-05
WIP: add sklmodel macro, add linear_model and ensemble mods

ZacLN:
Adds the `sklmodel` macro that adds default constructors, `clean!`, fit and predict functions for each model. Default parameters are taken from the web-scrape or scikit learn website but probably need a sense check by someone.

helper functions/traits etc will need to be added by hand

I think tests need to be added by someone a bit more familiar with these models.

PR #39, opened 2019-08-31
Sk-Learn integration (2)

tlienart:
This is a cleaning up of the PR #26 using the macro defined there, fixing the tests where required and adding metadata where required.

**Notes**

- Classifiers (_on hold for now --> will be for another PR_)
- did not add GraphicalLassoCV and RANSAC for now


ablaom:
Not reviewed in detail but looking great, thanks!

tlienart:
insignificant failure due to usage of `isnothing` which works with 1.1+ but not 1.0; will change for `===nothing`; otherwise currently pass.

tlienart:
Summary of the PR (now ready to merge):

all models listed below have

* [x] a constructor with default parameters and a reasonable "clean" method
* [x] been tested for their "fit" and "predict" method in a simple case (assumption is that the algorithm is correct, just checking that it "works")
* [x] a dedicated `fitted_params` which corresponds to the attributes of the fitresult as per sklearn docs; these have also been tested (checking that they're effectively there and have appropriate fields).
* [x] basic metadata corresponding to the package + metadata corresponding to the `input_scitype` and `target_scitype` (pretty much everything takes continuous and outputs continuous apart from the few multi-task ones which output multi-continuous)

### Models

**Linear regressors**: _all models_ (apart from RANSAC which should probably be considered an ensemble model)

**Gaussian Processes**: the GP regressor

**Ensemble models**: Adaboost, Bagging, GB and RF regressors

Issue #45, opened 2019-09-10
Target types for MultiTaskLassoCVRegressor, and so forth


ablaom:
@tlienart On the current dev branch:

```julia
julia> X, y = @load_boston;
julia> Y = (y1=y, y2=y)  # a Tables.jl compatible table
julia> fit!(machine(@load(MultiTaskLassoCVRegressor), X, Y)
```
will throw an error because  Y is not a matrix. Can we require multi-targets be tables? https://github.com/alan-turing-institute/MLJModels.jl/issues/40.  That is, in fact, what the `target_scitype` trait for this model reflects,  which is why we get a warning, if we use a matrix instead:

```julia
julia> X, y = @load_boston;
julia> fit!(machine(@load(MultiTaskLassoCVRegressor), X, hcat(y,y)))
┌ Warning: The scitype of `y`, in `machine(model, X, y)` is incompatible with `model`:
│ scitype(y) = AbstractArray{Continuous,2}
│ target_scitype(model) = ScientificTypes.Table{#s13} where #s13<:(AbstractArray{#s12,1} where #s12<:Continuous). 
└ @ MLJ ~/Dropbox/Julia7/MLJ/MLJ/src/machines.jl:43
[ Info: Training Machine{MultiTaskLassoCVRegressor} @ 3…51.
Machine{MultiTaskLassoCVRegressor} @ 3…51
```
Also, predictions are presently matrices, and not tables. 

Allowing tabular input means tweaking the macro `@sk_model` for multi-target predictors. A table comes in, which is coerced to a matrix for the internal fit method, using `MLJ.matrix` (which is a no-op if you actually give it a table. The internal predict's return value (a matrix) can be materialised as a table using `MLJ.table`, which adds dummy labels if no `names` are provided - probably just fine for now. A complete solution would bundle into the `fitresult` the column names for inclusion with `MLJ.table(matrix_predictions, names=names)`.

I don't think we should allow both matrices and tables, because this will complicate the loss functions API for multi-targets.

PR #46, opened 2019-09-10
addresses issue #45

tlienart:
Small tweak in the @sk_model macro so that

1. in the multi-target case, `fit` expects a table for `y` and `predict` outputs a table
2. the name of the targets are passed

**note**: here, like in GLM, there is this slightly clunky way of passing additional informations along the "actual" fitresult which is used by the external package for the prediction; I believe it's fine for now but in some future it may be relevant to have `fit` return a 4-tuple with

```julia
return fitresult, aux_infos, cache, report
```

where `aux_infos` is possibly `nothing` or stuff like target names etc which would be passed along `fitresult` to `predict`. (we can discuss this some other time)

### DIFF
--- a/src/ScikitLearn/ScikitLearn.jl
+++ b/src/ScikitLearn/ScikitLearn.jl
@@ -3,6 +3,7 @@ module ScikitLearn_
 #> for all Supervised models:
 import MLJBase
 using ScientificTypes
+using Tables
 
 #> for all classifiers:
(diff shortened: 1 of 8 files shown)


ablaom:
Awesome. This was evidently quite tedious to fix. 

Yeah, we need an `aux_info`. 

Many thanks.

PR #9, opened 2019-03-01
Adding KNN Classifier

kirtsar:
First try. I haven't figure out already about output kinds.

PR #47, opened 2019-09-11
Interface with NearestNeighbors.jl

tlienart:
And we should probably close the pr #9 

The test is a bit silly but it was mostly to check that dimensions are fine etc. 

Note the use of `permutedims` which will create a copy, transposes are not allowed to be passed to NNTree; there's an issue for this https://github.com/KristofferC/NearestNeighbors.jl/issues/82 where it's mentioned that materializing the transpose will lead to better performance.

### DIFF
--- a/src/Clustering.jl
+++ b/src/Clustering.jl
@@ -58,10 +58,7 @@ function MLJBase.fit(model::KMeans
 
     Xarray = MLJBase.matrix(X)
 
-    # NOTE see https://github.com/JuliaStats/Clustering.jl/issues/136
-    # this has been updated but only on #master, in the future replace permutedims
-    # with transpose (lazy)
-    result    = C.kmeans(permutedims(Xarray), model.k; distance=model.metric)
+    result    = C.kmeans(transpose(Xarray), model.k; distance=model.metric)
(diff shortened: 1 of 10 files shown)


src/Clustering.jl
```diff
@@ -58,10 +58,7 @@ function MLJBase.fit(model::KMeans
 
     Xarray = MLJBase.matrix(X)
```
ablaom: Option: you can now do `MLJBase.matrix(X, transpose=true)`

src/NearestNeighbors.jl
```diff
@@ -0,0 +1,134 @@
...
+            probas ./= sum(probas)
+        end
+        preds[i] = MLJBase.UnivariateFinite(classes, probas)
+    end
+    return preds
+end
+
+function MLJBase.predict(m::KNNRegressor, (tree, y), X)
+    Xmatrix     = permutedims(MLJBase.matrix(X))
+    idxs, dists = NN.knn(tree, Xmatrix, m.k)
```
ablaom: Maybe `MLJBase.matrix(X, transpose=true)`

Issue #56, opened 2019-09-17
Move model_utils.jl and parameter_utils.jl to MLJBase


ablaom:
These great tools should be available to anyone implementing the MLJ model interface.

tlienart:
Cool! Should I open a PR with these things in MLJBase? if so do you have a preference as to where the code ends up?

ablaom:
Sure, that'd be great. 

Could be added to MLJBase/src/utilities.jl but feel free to break this up if you want. Note that  `@set_defaults` is already there and used in MLJ/src/composites.jl. It might make sense to replace its use with the newer macro, but I would say that is quite low priority - if it ain't broke don't fix it. Still, your call.

I'm assuming you'd be happy to also make the necessary PR on MLJModels#dev to keep #dev in synch?

tlienart:
I'll do this after #60 is merged

ablaom:
*Todo*

[  ] Create migration PR for  MLJBase#dev (additions)

[  ] Create migration PR for MLJModels#dev (deletions and new re-export calls)

[  ] Create documentation PR on MLJ#dev for documentation of utilities in docs/src/adding_new_models_for_general_use.md

tlienart:
opened an issue for the last task  https://github.com/alan-turing-institute/MLJ.jl/issues/241

PR #58, opened 2019-09-17
Defmacros

tlienart:
This generally improves the macros `@mlj_model` and `sk_model` by

* extracting the different bits into _functions_ which are now called within the macros, this makes it a lot easier to reason about the macros and to maintain them
* macros are now reduced to the bare minimum calling those functions, this means that both macros are now  "in synch" (they call the same functions initially) which is good for maintenance

It now becomes easier to envisage extending the `@sk_model` macro for classifiers. In much the same way as the functions `_skmodel_predict` is defined, there would be a `_skmodel_predict_proba`. One of the two would be loaded conditional on whether the expression contains a `<: Probabilistic` or `<: Deterministic`. This is wip on branch `skc`.

Anyway I wanted to get this through this before any work to migrate code to MLJBase was done (#56) 

I don't think an in-depth review is required but I'll leave this open until tomorrow in case there are comments and will otherwise merge.

### DIFF
--- a/src/ScikitLearn/ScikitLearn.jl
+++ b/src/ScikitLearn/ScikitLearn.jl
@@ -22,7 +22,57 @@ import ..ScikitLearn
 
 include("svm.jl")
 
-import .._unpack!
+import .._process_model_def, .._model_constructor, .._model_cleaner
+
+"""
+_skmodel_fit
+
+Called as part of [`@sk_model`](@ref), returns the expression corresponing to the `fit` method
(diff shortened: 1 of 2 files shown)


tlienart:
The red cross is a red-herring, there was a failed HTTP connection on the MacOS with 1.2 instance but everything passes on the branch so all good.

ablaom:
This all makes sense to me, thanks. Merge when ready.

Issue #57, opened 2019-09-17
Sklearn models part 2


tlienart:
This issue is more to keep track of things than to ask for help

### Generalized Linear Models

https://scikit-learn.org/stable/modules/classes.html#module-sklearn.linear_model

?   | Name                    | Has predict_proba? | comment
--- | ----------------------- | ------------------ | ------------------------------
✅ | ARDRegression           | no                 |
✅ | BayesianRidge           | no                 |
✅ | ElasticNet              | no                 | 1
✅ | ElasticNetCV            | no                 | 1
✅ | HuberRegressor          | no                 |
✅ | Lars                    | no                 | 1
✅ | LarsCV                  | no                 | 1
✅ | Lasso                   | no                 | 1
✅ | LassoCV                 | no                 | 1
✅ | LassoLars               | no                 | 1
✅ | LassoLarsCV             | no                 | 1
✅ | LassoLarsIC             | no                 | 1
✅ | LinearRegression        | no                 |
❌ | LogisticRegression      | yes                | 2
❌ | LogisticRegressionCV    | yes                | 2
✅ | MultiTaskLasso          | no                 | 1
✅ | MultiTaskElasticNet     | no                 | 1
✅ | MultiTaskLassoCV        | no                 | 1
✅ | MultiTaskElasticNetCV   | no                 | 1
✅ | OrthogonalMatchingPursuit   | no             |
✅ | OrthogonalMatchingPursuitCV | no             |
❌ | PassiveAggressiveClassifier | no             |
✅ | PassiveAggressiveRegressor  | no             |
❌ | Perceptron              | no                 |
❌ | RANSACRegressor         | no                 |
✅ | Ridge                   | no                 |         |
❌ | RidgeClassifier         | no                 |
❌ | RidgeClassifierCV       | no                 |
✅ | RidgeCV                 | no                 |
❌ | SGDClassifier           | yes                |
✅ | SGDRegressor            | no                 |
✅ | TheilSenRegressor       | no                 |
     
1. could extract path
2. there's also predict_log_proba which we may want to interface with (?)

tlienart:
Hmm there's also LDA and QDA I should add

* [ ] [QDA](https://scikit-learn.org/stable/modules/generated/sklearn.discriminant_analysis.QuadraticDiscriminantAnalysis.html)
[...]

PR #60, opened 2019-09-19
Sklearn classifiers

tlienart:
See also #57 

@ablaom given that it's a massive PR, I guess reviewing it in details would be a big pain in the ass. Maybe look at the following couple of things:

1. the macro `@sk_model` and the way Probabilistic / Nonprob is handled + how decode is passed around, nothing too fancy
2. maybe one arbitrary example like RFClassifier, all of them follow the same pattern anyway, what's time consuming is to encode the constraints etc.

As for the other sklearn PR, there's always the risk that a constraint has not been well specified which would cause a bug in a given usecase where the user would specify the kwargs in a given way; I don't think there will be very many of those errors as I've spent a fair bit of time double checking but with so many models there may still be the odd one. 
Should be easy to fix once reported; I guess if we have hackathon with tolerant users where they'd try random models with their own inputs these things will appear.

### Adds

* Perceptron
* DummyClassifier
* LogisticRegression
* LogisticRegressionCV
* RidgeClassifier
* RidgeClassifierCV
* SGDClassifier both probabilistic and deterministic
* GPClassifier
* AdaboostClassifier
* BaggingClassifier
* GradientBoostingClassifier
* RandomForestClassifier
* BernoulliNB
* GaussianNB
* MultinomialNB
* ComplementNB
* KNeighborsClassifier
* KNeighborsRegressor
* RANSACRegressor
* XTreesClassifier and Regressor

There's probably something like a dozen other supervised models I haven't yet done but will be for another PR:

* **ensembles**: voting, ~~extratree~~, HGB (I may have had issues with some)
* ~~**NB**: BernoulliNB~~
* **Neighbors**: ~~KNeighborsClf and KNeighborsReg~~, Radius, Centroid, but NearestNeighbors.jl is already nice
* **neural net**: MLP reg and classif (also MLJFlux)
* **GM**: bayesian GM, GM
* **exotic**: kernel ridge, isotonic, calibrated, ...


src/Clustering.jl
```diff
@@ -5,8 +5,6 @@
 
 module Clustering_
 
-export KMeans, KMedoids
```
tlienart: FYI these are not needed anymore, the `@mlj_model` macro does the export.

ablaom: Ok.

src/ScikitLearn/ScikitLearn.jl
```diff
@@ -37,17 +40,24 @@ function _skmodel_fit(modelname, params)
...
+			y1 		  = nothing
+			# in multi-target regression case
 			if Tables.istable(y)
 			   yplain    = MLJBase.matrix(y)
 			   targnames = MLJBase.schema(y).names
 			end
+			if eltype(y) <: CategoricalString
+				yplain = MLJBase.int(y)
+				y1     = y[1]
+			end
```
ablaom: You need  `if eltype(y) <: Union{CategoricalString,CategoricalValue}` here.  But I am wondering why you need to test  machine types at all here. Wouldn't it be safer/cleaner to have separate `_skmodel_fit_classifier` method for classifiers?

src/ScikitLearn/exotic.jl
```diff
@@ -0,0 +1,28 @@
+# | model                  | build  | fitted_params | report | metadata | tests 1 | tests 2 |
+# | ---------------------- | ------ | ------------- | ------ | -------- | ------- | ------- |
+# | DummyClassifier        | ✓      | ✓             | ✗      | ✓        |  ✓      | ✓       |
+# | DummyRegressor         | ✗      | ✗             | ✗      | ✗        |  ✗      | ✗       |
+# | IsotonicRegressor      | ✗      | ✗             | ✗      | ✗        |  ✗      | ✗       |
+# | KernelRidgeRegressor   | ✗      | ✗             | ✗      | ✗        |  ✗      | ✗       |
+# | BGMClassifier          | ✗      | ✗             | ✗      | ✗        |  ✗      | ✗       |
+# | GMClassifier           | ✗      | ✗             | ✗      | ✗        |  ✗      | ✗       |
+
+DummyClassifier_ = ((ScikitLearn.Skcore).pyimport("sklearn.dummy")).DummyClassifier
```
ablaom: Could you explain a bit what this line if for?

ablaom:
Okay, I can see a lot of work and good thinking has gone into this, many thanks. Thanks also for the review guidance; that was helpful.

- I wonder if we could add something a bit meatier to `simple_test_classif_prob` and
  `simple_test_classif` to check that that at least the coding/encoding is working properly, ie that the           
  levels are not getting inadvertantly scrambled. 

- Some classifier have a hyperparameter such as `multi_class` to tell the model class information. For example, LogisticClassifier has this. As I understand it, this information is actually redundant as we should be taking care of this under the hood, no? In that case, the hyperparameter should either be removed (might lead to nasty case distinctions) or restricted to a fixed value ("auto" in this case).  What's your take?

- It seems that all classifiers have `target_scitype=AbstractVector{<:Finite}`. Are there no classifiers that do only binary classification - in which case we should have `target_scitype=Abstract{<:Binary}` (`Binary` being alias for `Finite{2}`)?

src/ScikitLearn/exotic.jl
tlienart: These (and all other `model_`) get used in the macro, specifically this line which is in the `_skmodel_fit`  function in `MLJModels/src/ScikitLearn/ScikitLearn.jl`:

```julia
skmodel = $(Symbol(modelname, "_"))($((Expr(:kw, p, :(model.$p)) for p in params)...))
```

Basically the machinery is as follows:

* assign the python constructor to a name with an underscore : `Model_`
* declare a struct with name  `Model`
* automatically define a function `fit` which will internally  call `Model_(...)` which will in turn use the python backend.

src/ScikitLearn/ScikitLearn.jl
tlienart: changed to the union, thanks.

With respect to your proposal, it's a bit tricky: at macro time we don't know if it's a classifier or a regressor (we could read the  model name but that's hacky); so we could not have a `_skmodel_fit_classifier`; what we could have is that this function is duplicated something like:

```julia
function _skmodel_fit(...)
    quote
        function MLJBase.fit(model::$modelname, verbosity::Int, X, y::AbstractVector{<:Union{CategoricalString,CategoricalValue}}
         ....
        end
        function MLJBase.fit(..., X, y)
        ....
        end
    end
end
```

However this has several issues:

1. it creates (a lot of) methods which will never be used
2. it forces `y` to be an AbstractVector, I prefer `eltype` because it works on anything

So I think in this case it may be better to stick with this.

tlienart:
Ok great, with respect to your  three comments:

**first comment** _improving the simple tests_:

this is already checked as far as I'm aware with `@test Set(unique(p)) == Set(unique(y))`; no doubt these `simple_test_*` will be reinforced as we go; I just thought I'd get the ball started to make testing of many models more efficient

**second and third comment** _multi class etc_

So scikit learn makes any binary classifier into multi-class by using OVR; in some cases you can specify whether you want to use OVR or something else (e.g. if there's a specific loss function) in which case there's the hyper parameter `multi_class`. 
As an example, the logistic regresssion has that `multi_class` parameter: if you specify  `multi_class="multinomial"` then it will train using the multinomial loss function (which not supported by all the solvers they offer, in particular not by `liblinear`) whereas in the OVR it will use the binary loss multiple times. They have these options because not all solvers support all modes. In that case the `"auto"` mode will use `OVR` if you specify the solver to be `liblinear` and multinomial otherwise.  
I suggest we leave this untouched as, for the moment, MLJ does not yet support default OVR for binary classifiers as far as I'm aware. In the future if we have OVR and it's somehow better than what SkLearn does internally then we could encourage users to use that but I don't think it's in our advantage to push this now as it would effectively reduce the number of models we have.

I think this answers the last comment as well.

tlienart:
The changes are somewhat irrelevant, just a few more models added and addressed the `<:CategoricalValue` comment. Ignore the red crosses, sometimes tests fail because they fail to clone Conda properly so I had to stop a bunch of them.

src/ScikitLearn/ScikitLearn.jl
ablaom: Sorry I don't understand. Does not the caller of `@sk_model` know whether it wants a regressor or classifier? We could have `@sk_model_regressor` and `@sk_model_classifier`, no?

src/ScikitLearn/exotic.jl
ablaom: Ah, I see. Thanks!

ablaom:
> this is already checked as far as I'm aware with @test Set(unique(p)) == Set(unique(y)); no doubt these simple_test_* will be reinforced as we go; I just thought I'd get the ball started to make testing of many models more efficient

No, I mean  scrambled in sense `male` could come out `female` and vice-versa. Maybe since  they're  all essentially the same we could just test DummyClassifier. So, maybe do `y = categorical(rand("abcd", 100))`, add a zero-count level `d` for good measure (using `levels!(y, collect("abcd"))`),  compute the frequencies with `StatsBase.countmap(y)` and compare with the predicted distribution (using `broadcast(pdf, d, MLJBase.classes(d))`, after checking `classes(d) == collect("abcd")`). I got caught out with this in wrapping classifiers before and would feel happier with this one test. 

Thanks for clarifying the multi-class thing. I agree we just leave it then.

tlienart:
Ok here's what I've done:

- added the generation of two super dummy classifier examples with clouds of point that are very separated and testing points that are in those clouds
- then checking that the predictions get at least 75% accuracy (which means that the labelling happens correctly btw). 

I've done this both for 3 classes and 2 classes; using the 3-class case mostly and the 2-class case for crap classifiers (like dummyclassifier) that struggle otherwise. 

For all probabilistic classifiers, the 3 class case works perfectly fine, for the few deterministic classifiers, performances are much worse so I used the 2 class case then. 

Note that I don't think this lack of performance could be related to the encoding in the deterministic case; the encoding as it happens for deterministic classifier is identical to that of SVM models you implemented and I doubt if the encoding was wrong you'd get over 75%  accuracy in the 2 class case.

tlienart:
duplicate macro for `@sk_model_regressor` and `@sk_model_classifier` so that we don't have to test the type of `y[1]`.

tlienart:
Separated `@sk_reg` and `@sk_clf`; will wait for tests to go green and merge.

PR #75, opened 2019-10-04
Discretizer

vollmersj:
a port from koala


vollmersj:
Ported discretiser from Koala

src/builtins/Transformers.jl
```diff
@@ -177,6 +179,91 @@ MLJBase.docstring(::Type{<:FeatureSelector})      = "Filter features (columns) o
 MLJBase.load_path(::Type{<:FeatureSelector})      = "MLJModels.FeatureSelector"
 
 
+
+#### UNIVARIATE Discretizer ####
+"""
+UnivariateDiscretizer(n_classes=512)
+Returns a `MLJModels` for discretizing vectors of `Real`
+eltype, where `n_classes` describes the resolution of the
```
ablaom: - `MLJModel` not `MLJModels`.

- Can we instead make this "for discretising any `Continuous` vector `v` (`scitype(v) <: AbstractVector{Continuous}`)" ?

src/builtins/Transformers.jl
```diff
@@ -177,6 +179,91 @@ MLJBase.docstring(::Type{<:FeatureSelector})      = "Filter features (columns) o
...
+discretization. Transformed vectors are of eltype `Int46`. The
+transformation is chosen so that the vector on which the transformer
+is fit has, in transformed form, an approximately uniform distribution
+of values.
+### Example
+    using MLJ
+    using MLJModels
+    t = UnivariateDiscretizer(n_classes=10)
```
ablaom: Can delete `using MLJModels`; it is superfluous.

src/builtins/Transformers.jl
```diff
@@ -177,6 +179,91 @@ MLJBase.docstring(::Type{<:FeatureSelector})      = "Filter features (columns) o
...
 
 
+
+#### UNIVARIATE Discretizer ####
+"""
+UnivariateDiscretizer(n_classes=512)
+Returns a `MLJModels` for discretizing vectors of `Real`
+eltype, where `n_classes` describes the resolution of the
+discretization. Transformed vectors are of eltype `Int46`. The
+transformation is chosen so that the vector on which the transformer
```
ablaom: I really think we should make the output be a `CategoricalVector`, not an `Int` vector. So we are converting `Continuous` to `OrderedFactor`. 

Here's how you do that:

If `w` is the integer vector, return `categorical(w, ordered=true)`. For inverse transforming a categorical vector `w_cat`, you can first get the integer version using `w=MLJBase.int(w_cat)` and then apply the existing inverse transform to get the final continuous vector.

test/featureselecturrule.jl
```diff
@@ -0,0 +1,45 @@
+using MLJBase
```
ablaom: I don't think you mean this to be in the current PR but part of a new PR, no?

src/builtins/Transformers.jl
```diff
@@ -177,6 +179,91 @@ MLJBase.docstring(::Type{<:FeatureSelector})      = "Filter features (columns) o
...
+                           w::AbstractVector{T}) where T<:Integer
+    return [inverse_transform(transformer, result, k) for k in w]
+end
+
+
+MLJBase.input_scitype(::Type{<:UnivariateDiscretizer})  = AbstractVector{<:MLJBase.Infinite}
+MLJBase.output_scitype(::Type{<:UnivariateDiscretizer}) = AbstractVector{MLJBase.Continuous}
+MLJBase.docstring(::Type{<:UnivariateDiscretizer})      = "Discretise continuous variables via quantiles"
+MLJBase.load_path(::Type{<:UnivariateDiscretizer})      = "MLJModels.UnivariateDiscretizer"
+
```
ablaom:
```julia
MLJBase.input_scitype(::Type{<:UnivariateDiscretizer})  = AbstractVector{<:MLJBase.Continuous}
MLJBase.output_scitype(::Type{<:UnivariateDiscretizer}) = AbstractVector{<:MLJBase.OrderedFactor}
```

src/builtins/Transformers.jl
```diff
@@ -177,6 +179,171 @@ MLJBase.docstring(::Type{<:FeatureSelector})      = "Filter features (columns) o
 MLJBase.load_path(::Type{<:FeatureSelector})      = "MLJModels.FeatureSelector"
 
 
+mutable struct ToIntTransformer <: Unsupervised
```
tlienart: please add a docstring

src/builtins/Transformers.jl
```diff
@@ -177,6 +179,171 @@ MLJBase.docstring(::Type{<:FeatureSelector})      = "Filter features (columns) o
...
+    sorted::Bool
+    initial_label::Int # ususally 0 or 1
+    map_unseen_to_minus_one::Bool # unseen inputs are transformed to -1
+end
+
+ToIntTransformer(; sorted=false, initial_label=1,
+                 map_unseen_to_minus_one=false) =
+                     ToIntTransformer(sorted, initial_label,
+                                      map_unseen_to_minus_one)
+
```
tlienart: please make this a function so that it doesn't go over 4 lines

src/builtins/Transformers.jl
```diff
@@ -177,6 +179,171 @@ MLJBase.docstring(::Type{<:FeatureSelector})      = "Filter features (columns) o
...
+        if isa(exception, KeyError)
+            if transformer.map_unseen_to_minus_one
+                ret = -1
+            else
+                throw(exception)
+            end
+        end
+    end
+    return ret
+end
```
tlienart: Please use the `get ... do ` syntax, e.g.

```julia
a = Dict("a"=>0)
b = get(a, "b") do
  nothing # or the error or whatever
end
```

don't use try catch if you can avoid it

src/builtins/Transformers.jl
```diff
@@ -177,6 +179,171 @@ MLJBase.docstring(::Type{<:FeatureSelector})      = "Filter features (columns) o
...
+MLJBase.inverse_transform(transformer::ToIntTransformer, result, y::Int) = result.T_given_int[y]
+
+# vector case:
+function MLJBase.transform(transformer::ToIntTransformer, result::ToIntResult{T},
+                   v::AbstractVector{T}) where T
+    return Int[transform(transformer, result, x) for x in v]
```
tlienart: Shouldn't this be `Int.(...)` ?

src/builtins/Transformers.jl
```diff
@@ -177,6 +179,171 @@ MLJBase.docstring(::Type{<:FeatureSelector})      = "Filter features (columns) o
...
+end
+MLJBase.inverse_transform(transformer::ToIntTransformer, result::ToIntResult{T},
+                  w::AbstractVector{Int}) where T = T[result.T_given_int[y] for y in w]
+
+MLJBase.input_scitype(::Type{<:ToIntTransformer})  = AbstractVector{<:MLJBase.Infinite}
+MLJBase.output_scitype(::Type{<:ToIntTransformer}) = AbstractVector{MLJBase.Count}
+MLJBase.docstring(::Type{<:ToIntTransformer})      = "transforms values to Int"
```
tlienart: `transforms` -> `Transforms`

Also probably better to use the `metadata_model` please

src/builtins/Transformers.jl
```diff
@@ -177,6 +179,171 @@ MLJBase.docstring(::Type{<:FeatureSelector})      = "Filter features (columns) o
...
+    return UnivariateDiscretizerResult(odd_quantiles, even_quantiles), cache, report
+end
+# transforming scalars:
+function MLJBase.transform(transformer::UnivariateDiscretizer, result, r::Real)
+    k = 1
+    for level in result.odd_quantiles
+        if r > level
+            k = k + 1
+        end
+    end
```
tlienart:
```julia
k = sum(r .> result.odd_quantiles)
```

test.jl
```diff
@@ -0,0 +1,45 @@
+using MLJBase
```
tlienart: remove this file

ablaom:
@tlienart As I stated above, none of the `ToIntTransformer` code (ported from Koala) is relevant in the MLJ setting and could be dumped. I think removing this should resolve most of your comments above.

ablaom:
@vollmersj Can you remove the ToIntTransformer code?

vollmersj:
done

tlienart:
there's still a few comments that apply I reckon

ablaom:
Re: the travis fail. There is no need to annotate the type of `v` here:

```julia
function MLJBase.transform(transformer::UnivariateDiscretizer, result,
                  v) where (scitype(v) <: AbstractVector{Continuous})
   w=[transform(transformer, result, r) for r in v]
   return categorical(w, ordered=true)
end
```

Your trait declaration

```julia
MLJBase.input_scitype(::Type{<:UnivariateDiscretizer})  = AbstractVector{<:MLJBase.Continuous}
```

ensures that you method is never called with the wrong type.

ablaom:
I have added some stuff to ensure the output of the discretizer has elements with the same categorical pool every time, as well as some helper functions for dealing with categorical data that should be moved to MLJBase at some point.

ablaom:
@vollmersj Many thanks for your work on this!

Issue #77, opened 2019-10-08
DecisionTree.jl v 0.9.0 is incompatible with MLJModels


ablaom:
From this travis build, https://travis-ci.com/alan-turing-institute/MLJModels.jl/jobs/240146086, I get (near the end):

```julia
 LoadError: ArgumentError: UnivariateFinite: the condition Dist.isprobvec(p) is not satisfied.
4634  Stacktrace:
4635   [1] macro expansion at /home/travis/.julia/packages/Distributions/wRw5p/src/utils.jl:6 [inlined]
4636   [2] MLJBase.UnivariateFinite(::OrderedCollections.LittleDict{CategoricalArrays.CategoricalString{UInt8},Float64,Array{CategoricalArrays.CategoricalString{UInt8},1},Array{Float64,1}}) at /home/travis/.julia/packages/MLJBase/t7rZM/src/distributions.jl:102
4637   [3] MLJBase.UnivariateFinite(::Array{CategoricalArrays.CategoricalString{UInt8},1}, ::Array{Float64,1}) at /home/travis/.julia/packages/MLJBase/t7rZM/src/distributions.jl:119
4638   [4] #4 at ./none:0 [inlined]
4639   [5] iterate at ./generator.jl:47 [inlined]
4640   [6] collect_to!(::Array{MLJBase.UnivariateFinite{String,UInt8,Float64},1}, ::Base.Generator{UnitRange{Int64},getfield(MLJModels.DecisionTree_, Symbol("##4#5")){MLJModels.DecisionTree_.DecisionTreeClassifier,Array{CategoricalArrays.CategoricalString{UInt8},1},Array{Float64,2}}}, ::Int64, ::Int64) at ./array.jl:656
4641   [7] collect_to_with_first! at ./array.jl:643 [inlined]
4642   [8] collect(::Base.Generator{UnitRange{Int64},getfield(MLJModels.DecisionTree_, Symbol("##4#5")){MLJModels.DecisionTree_.DecisionTreeClassifier,Array{CategoricalArrays.CategoricalString{UInt8},1},Array{Float64,2}}}) at ./array.jl:624
[...]

PR #80, opened 2019-10-09
Resolve DecistionTree incompatibility (#77)

ablaom:
It seems passing categorical elements into DecisionTreeClassfier causes problems with the new Multithreaded version. Instead we now pass integer-encoded targets in the way we do in most other classifiers, which is more performant anyhow. The only downside is that the "pretty tree" printouts you get by setting verbosity>1 refer to encoded levels not original levels. To address this we now return the encoding as part of the output of `fitted_params`.


ablaom:
Okay after looking at these tests, I suspect a bug in DecistionTree.jl 0.9.0.

I'm leaving this alone for now.

ablaom:
https://github.com/bensadeghi/DecisionTree.jl/issues/105

PR #87, opened 2019-10-15
Featureselectordev

vollmersj:
The big question is if we should export individual feature rules like this

```
const StdSelector = MLJModels.FeatureSelectorRule{StdRule}
MLJBase.load_path(::Type{<:FeatureSelector})   = "MLJModels.StdSelector"
MLJBase.package_url(::Type{<:FeatureSelector}) = "https://github.com/alan-turing-institute/MLJModels.jl"
MLJBase.package_license(::Type{<:FeatureSelector}) = "MIT"
MLJBase.package_name(::Type{<:FeatureSelector})    = "MLJModels"
MLJBase.package_uuid(::Type{<:FeatureSelector})    = "d491faf4-2d78-11e9-2867-c94bc002c0b7"
MLJBase.is_pure_julia(::Type{<:FeatureSelector})   = true
MLJBase.input_scitype(::Type{<:FeatureSelector})   = MLJBase.Table(Scientific) # anything goes
MLJBase.output_scitype(::Type{<:FeatureSelector})  = MLJBase.Table(Scientific)
```


src/builtins/Transformers.jl
```diff
@@ -182,35 +182,52 @@ MLJBase.load_path(::Type{<:FeatureSelector})      = "MLJModels.FeatureSelector"
...
     features=[sch.names[m] for m in mask]
 
     transformer.fs=FeatureSelector(features)
     return fit(transformer.fs,verbosity,X)
 end
 
-MLJBase.fitted_params(::FeatureSelectorRule, fitresult) = (features_to_keep=fitresult,)
-
-function transform(transformer::FeatureSelectorRule,features, X)
+function MLJBase.transform(transformer::FeatureSelectorRule{R},features, X) where R<:SelectorRule
```
tlienart: spaces

src/builtins/Transformers.jl
```diff
@@ -182,35 +182,52 @@ MLJBase.load_path(::Type{<:FeatureSelector})      = "MLJModels.FeatureSelector"
...
     return transform(transformer.fs,features,X)
```
tlienart: here too (and everywhere)

src/builtins/Transformers.jl
```diff
@@ -182,35 +182,52 @@ MLJBase.load_path(::Type{<:FeatureSelector})      = "MLJModels.FeatureSelector"
...
 end
 
 
 
+MLJBase.fitted_params(transformer::FeatureSelectorRule{R}, fitresult) where R<:SelectorRule = (features_to_keep=fitresult,)
```
tlienart: typing is not needed
```julia
MLJBase.fitted_params(transformer::FeatureSelectorRule, fr) = (features_to_keep = fr,)
```

test.jl
```diff
@@ -0,0 +1,45 @@
+using MLJBase
```
tlienart: Please do not put this in the top level, it should go. in `test/` subfolder, and give it an appropriate name (put it with the builtins)

test.jl
```diff
@@ -0,0 +1,45 @@
...
+
+sch = MLJBase.schema(X)
+fsr=MLJModels.FeatureSelectorRule(rule=var_rule)
+params(fsr)
+fitresult, _, report = fit(atom_ols, 1, Xtrain, ytrain)
+pipe=@pipeline MyPipe(fsr=fsr, ols=atom_ols)
+
+TunedModel(model=pipe, tuning=Grid(), resampling=Holdout(),
+measure=rms, operation=predict, ranges=range(fsr,:kwargs,[(:h=>8.0,),(:h=>9.0,)]),
+minimize=true, full_report=true)
```
tlienart: please add some actual tests in this ...  (or remove this file if you didn't mean to push it)

test/Transformers.jl
```diff
@@ -43,27 +42,30 @@ infos = info_dict(selector)
...
-    return s <: Continuous ? (std(X[n])>0.1 ? true : false) : true
+mutable struct StdRule <: MLJModels.SelectorRule
+     threshold::Float64
 end
+(sr::StdRule)(X,name,type,scitypes) =  scitypes <: Continuous ? (std(X[name])>sr.threshold ? true : false) : true
+const StdSelector = MLJModels.FeatureSelectorRule{StdRule}
+StdSelector(;threshold=0.2)= StdSelector(StdRule(threshold))
+fsr=StdSelector()
+fsr_fit,=fit(fsr,1,X)
+
```
tlienart: add tests for the rule and the structure of FeatureSelectorRule testing fields etc + also test creation of rules + what happens if your rule does not meet the specs, is there an appropriate error message? add one

test/featureselecturrule.jl
```diff
@@ -0,0 +1,63 @@
+using MLJBase
```
tlienart: the name of this file is incorrect and not at the right place + the whole test file should be a module like other test files, please look at `Transformer.jl` and either just put your code in there or imitate the style.

test/featureselecturrule.jl
```diff
@@ -0,0 +1,63 @@
...
+fitresult, _, report = fit(atom_ols, 1, Xtrain, ytrain)
+pipe=@pipeline MyPipe(fsr=fsr, ols=atom_ols) is_probabilistic=false
+fit(pipe,1,X,y)
+
+tm= TunedModel(model=pipe, tuning=Grid(), resampling=Holdout(),
+measure=rms, operation=predict, ranges=range(pipe,:(fsr.rule.threshold),lower=8.0,upper=9.0),
+minimize=true, full_report=true)
+fit(tm,1, X,y)
+params(pipe)
+params(pipe)
```
tlienart: make sure there are actual tests...

src/builtins/Transformers.jl
```diff
@@ -182,35 +182,52 @@ MLJBase.load_path(::Type{<:FeatureSelector})      = "MLJModels.FeatureSelector"
     FeatureSelectorRule(rule::Function)
 
 An unsupervised model for filtering features (columns) of a table.
-Only those features that satisfy the rule
-:name, :scitype, :type
+This a parmetric type that expects a callable mutable struct encapsulating
+    the parameters of the rule and the rule being returning true and false based
+    on (X,name,type,scitypes) where X is data treated as a table
```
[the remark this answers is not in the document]
tlienart: you also need to check that the rule is appropriate, if someone gives a wronog rule (does not meet the specs), it should be detected

ablaom:
I don't really understand why we are wrapping FeatureSelector here. Why not just add the hyperparameter `rule` to the existing `FeatureSelector` transformer, and give it the default rule that allows everything in the existing `features` field through?

A few notes:

- Every struct <: Model must be given a pure-keyword constructor with no reference to type parameters. This means default values must be chosen for all fields and any type parameters. So `FeatureSelectorRule()` must make sense. 

- In the current API there can only be one registry metadata entry per struct <:Model and you cannot assign aliases to model type names (as in `const MyFeatureSelector = FeatureSelector{Rule}`).  Sorry for misleading you earlier on this point. Changing this introduces complications to the API and model registry building process I prefer to avoid, and I have not seen sufficient examples to warrant doing so now. Instead of doing something like `PaintItBlack = PaintIt{:black}, PaintItWhite = PaintIt{:white}` to define aliases and generate separate registry entries (which won't work) you need to define separate structs `PaintItBlack` and `PaintItWhite`. You can avoid code duplication for the constructors, `transform` and other methods using standard dispatch tricks, as in:

```
paint_it_transform(model, color, fitresult, X) = ...
transform(model::PaintIt{color}, args...) = paint_it_transform(model, color, args...)
transform(model::PaintItBlack, args...) = paint_it_transform(model, :black, args...)
transform(model::PaintWhite, args...) = paint_it_transform(model, :white, args...)
```

Whether you actually want to alias multiple methods for your various rules is another question. But for the record, this is how I recommend it be done.

src/builtins/Transformers.jl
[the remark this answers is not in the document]
vollmersj: I can check the number of arguments of a function but other try and catch there is no other way to safely evaluate.

vollmersj:
Does this mean  does this mean you want
```
mutable struct PaintItBlack{
     PaintIt{:black}
}
```
There is probably a macro that makes this easer?

ablaom:
No. 

Let's say you have generic model `PaintIt{C}` one of the fields is
`color` whose type is `C`. Two common , choices for `color` are
`black` and `white`, which have different types `Black` and `White`
(unlike my first example above).

Let's say you want three models in
the registry, one generic `PaintIt` and two specialised versions
`PaintItBlack`, `PaintItWhite`. Then I'm suggesting:

```julia
mutable struct PaintIt{C}
    color::C
    brush::Int
end

mutable struct PaintItBlack
    brush::Int
end

mutable struct PaintItWhite
    brush::Int
end

# constructors:
PaintIt(; color=black, brush=0) = PaintIt(color, brush)
PaintItBlack(; brush=0) = PaintIt(black, brush)
PaintIt(; brush=0) = PaintIt(white, brush)

# transform method:
paint_it_transform(model, color, fitresult, X) = ... <lots of code>

transform(model::PaintIt{C}, args...) = paint_it_transform(model, C, args...)
transform(model::PaintItBlack, args...) = paint_it_transform(model, black, args...)
transform(model::PaintWhite, args...) = paint_it_transform(model, white, args...)

**Comment by ablaom** [comment]
Also, at most one set of metadata entries per model struct. So in the above example, you could have 

```julia
target_type(::Type{<:PaintIt}) = AbstractVector{<:Union{Continuous,Finite}}
target_type(::Type{<:PaintItBlack}) = AbstractVector{<:Finite}
target_type(::Type{<:PaintItWhite}) = AbstractVector{<:Continuous}
```
But adding
```julia
target_type(::Type{<:PaintIt{Orange}}) = AbstractVector{<:GreyImage}
```
will have no consequence for the general user.

vollmersj:
It seems I need to wrap the struct in another struct.

```
mutable struct StdFeatureSelector
    threshold::Float64
end
StdFeatureSelector(;threshold=1.0)=FeatureSelector(rule=StdRule(threshold))
S=StdFeatureSelector(threshold=0.2)
typeof(S) #FeatureSelector{StdRule} as opposed to StdFeatureSelector

```

ablaom:
Closing this very old PR to remove clutter. Feel free to re-open if revisiting.

Issue #64, opened 2019-09-24
Towards 0.5.2


ablaom:
- [x] (**Enhancement**) Update MLJBase to 0.7.0 to remove cap on CategoricalArrays and dependent packages, such as  DataFrames. 

- [x] (**Enhancement**) Remove compatability restriction on DecistionTree.jl, after testing (#80, https://github.com/bensadeghi/DecisionTree.jl/issues/105)

ablaom:
https://github.com/JuliaRegistries/General/pull/4448#issuecomment-542973199

PR #88, opened 2019-10-17
Towards 0.5.2

ablaom:
#64

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,12 +1,11 @@
 name = "MLJModels"
 uuid = "d491faf4-2d78-11e9-2867-c94bc002c0b7"
 authors = ["Anthony D. Blaom <anthony.blaom@gmail.com>"]
-version = "0.5.1"
+version = "0.5.2"
 
 [deps]
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
 ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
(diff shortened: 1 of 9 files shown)


DilumAluthge:
@ablaom @tlienart Out of curiosity, could you explain the workflow that you use with MLJ? Is it the git-flow workflow, or something different?

ablaom:
Thanks for the query @DilumAluthge, 

I guess it is git-flow, but I don't know how strict we're. Developers now make PR's to a dev branch and with small exceptions we only merge onto master from dev and only when when a new release is to be tagged. The exceptions are non-code stuff like readme and news updates. 

I have also pushed manual updates directly to master between releases, but I as these are actually code-generated, I'm inclined to stop doing that. To be honest, we are experimenting a bit with how we do things. Suggestions welcome. 

The biggest challenge is managing several closely integrated repos, especially when we decide code in one should be migrated to another. And when we make breaking changes to MLJBase, rolling out releases for rest of the eco-system has been painful. But now auto-merge is going to fix all that, yes ? 😄

DilumAluthge:
I don't have any suggestions to contribute, but thanks for sharing your workflow! I'm always looking to learn about new workflows to try.

DilumAluthge:
Just FYI, currently your `[compat]` entry for Tables is `Tables = "<0.1.19, >= 0.2"`. This does not have an upper bound, so your PR will not be automerged.

PR #99, opened 2019-10-27
Cleanup

### LINKED ISSUE #98: MultivariateStats predict is wrong
The test does not make sense and only predicts a single class; there's something incorrect in there.

The `cleanup` branch contains a bunch of fixes including problem with dimensions but the predit is still incorrect, WIP to fix it (my fault from merging it...)

tlienart:
This PR is a major cleanup effort primarily to:

- uniformize how the interfaces are structured
- use the `@mlj_model` macro where possible (makes it easier to maintain stuff if we use the same tools everywhere)
- use  the `metadata_*` where possible (same comment)
- general code review/improvements where possible
- update of the registry

### LDA work
- closes #98 removing Bayes' rule which shouldn't have been  there and using a softmax
- removes the CovarianceEstimation hard deps
- allows passing any covariance estimator `<: CovarianceEstimator` for both `cov_w` and `cov_b`
- improved docstring
- allows any semi metric to be used (not just SqEuclidean)
- sensible tests

(my fault for merging this too quickly)

### Future work

I haven't looked at everything in this pass (e.g. not at XGBoost or LIBSVM); but hopefully this sets the bases for a convergence in style for future interfaces etc.

### Merging

I think the changes here should be broadly non-controversial so I'll just merge this into dev granted tests pass


OkonSamuel:
Did you check with another dataset. The test predicts some downs as well. If you print out each member you would see this

OkonSamuel:
I used bayes rule here. Because there are times one would like to predict based on some probability threshold rather than just minimizing the misclassification rate.

OkonSamuel:
@tlienart does the predict method work now?

tlienart:
Thanks Sam; it was definitely not correct; a simple way to test it was to generate dummy data and check the misclassification rate which was 100%. Indeed the Bayesian rule caused most of the issue; if you look up the algo in the paper from Li et al; or in fact from Sklearn you’ll see that they use a soft max; there may be a confusion with how NaiveBayes works.

Anyway it’s all good now as far as I know :)

OkonSamuel:
Okay. Thanks. I will look at the paper. 
Maybe you should have left the model name as `MulticlassLDA` to separate it from the other in MultivariateStats package. `SubspaceLDA` which I plan on adding later . Or what do you think?

tlienart:
It would be great if you were to add subspace LDA! In my opinion given it’s a bit less standard it would be fine to call it SubspaceLDA and leave LDA as such; I changed the name because there were inconsistencies in how the different objects were called but it wasn’t a very strong opinion :)

OkonSamuel:
@tlienart. Sorry to drag this issue. I just read through the paper I have the following points to make. 
* they classified based on a distance measure in the transformed space. 
* the most popular measure used in most papers is mahalobi distance measure. 
* you are right sklearn uses softmax to predict. But it also uses class prior to generate discriminants first which is kinda bayes. 
* I was curious and carried out tests on the original bayes implementation using the tests you wrote and got misclassification rate of 0.088.
* I checked R mass library and also found that class prior were used in their code. 
* The bayes implementation is based off reduced rank lda in Elements of Statistical learning. Where they first reduced the dimension then generated posteriors using bayes rule.

tlienart:
I’m happy for you to suggest modifications to the current code including the Bayes prior! My only request would be to have tests that give similar results to comparable code in R (which is what you say); it may be me who screwed something up in the first place but at least I’m comfortable that the current tests and results make sense.

I think we agree with the rest; so maybe if you want to give this a shot is to re-incorporate bayes to the current code and check that the tests where I generate 3 separated clouds of points can be classified with a misclassification rate below 10-15%

OkonSamuel:
Okay. I'll get to work on that.

tlienart:
Thanks! Also having now read the stuff myself, I should be able to provide better comments for future work

ablaom:
@OkonSamuel Many thanks for you work and @tlienart for the review.

PR #101, opened 2019-10-29
Re-implemented LDA using Bayes

OkonSamuel:
This PR Re-implements LDA using bayes. Few things are worth nothing
* I edited the test for cross-entropy since the code gave smaller values.
* I removed the `metric` keyword as bayes classifier uses the Mahalanobis distance metric or equivalently spheres the data and uses the Euclidean metric
* Added `priors` keyword to enable users define prior probabilities as a `UnivariteFinite` distribution.


src/MultivariateStats.jl
```diff
@@ -271,8 +271,7 @@ $LDA_DESCR
 * `cov_b=SimpleCovariance()`: same as `cov_w` but for the between-class covariance.
 * `out_dim`: the output dimension, i.e dimension of the transformed space, automatically set if 0 is given (default).
 * `regcoef`: regularization coefficient. A positive value `regcoef * eigmax(Sw)` where `Sw` is the within-class covariance estimator, is added to the diagonal of Sw to improve numerical stability. This can be useful if using the standard covariance estimator.
-* `dist=SqEuclidean`: the distance metric to use when performing classification (to compare the distance between a new point and centroids in the transformed space), an alternative choice can be the `CosineDist`.
```
tlienart: This is not, per-se, required, so please leave as it was where it is a kwargs with the preferrential SqEuclidean.

src/MultivariateStats.jl
```diff
@@ -282,7 +281,7 @@ For more information about the algorithm, see the paper by Li, Zhu and Ogihara,
     cov_b::CovarianceEstimator = MS.SimpleCovariance()
     out_dim::Int     = 0::(_ ≥ 0)
     regcoef::Real    = 0.0::(_ ≥ 0)
-    dist::SemiMetric = SqEuclidean()
+    priors::Union{Nothing, MLJBase.UnivariateFinite} = nothing
```
tlienart: Let's not use union types if we don't need to, here nothing is the same as providing a UnivariateFinite with uniform probabilities on all classes

src/MultivariateStats.jl
```diff
@@ -292,7 +291,8 @@ function MLJBase.fit(model::LDA, ::Int, X, y)
     # NOTE: copy/transpose
     Xm_t   = MLJBase.matrix(X, transpose=true) # now p x n matrix
     yplain = MLJBase.int(y) # vector of n ints in {1,..., nclasses}
-    p      = size(Xm_t, 1)
+    N      = size(Xm_t, 2)  # Number of training samples
```
tlienart: please use `n`  (lowercase) (convention thing); also just do `(p, n) =  size(Xm_t)`

src/MultivariateStats.jl
```diff
@@ -301,37 +301,61 @@ function MLJBase.fit(model::LDA, ::Int, X, y)
     # check if the given one is sensible
     out_dim ≤ def_outdim || throw(ArgumentError("`out_dim` must not be larger than `min(p, nc-1)` where `p` is the dimension of `X` and `nc` is the number of classes."))
 
+    ## Estimates prior probabilities is unspecified by user.
+    if model.priors == nothing
```
tlienart: see comment above

src/MultivariateStats.jl
```diff
@@ -301,37 +301,61 @@ function MLJBase.fit(model::LDA, ::Int, X, y)
...
 end
 
-function MLJBase.predict(m::LDA, (core_res, class_list), Xnew)
+function MLJBase.predict(m::LDA, (core_res, class_list, priors), Xnew)
     # projection of Xnew XWt is n x o  where o = number of out dims
     XWt = MLJBase.matrix(Xnew) * core_res.proj
     # centroids in the transformed space, nc x o
     centroids = permutedims(core_res.pmeans)
 
+    N  = core_res.stats.tweight # N is the Number of training examples
```
tlienart: `n = `

src/MultivariateStats.jl
```diff
@@ -301,37 +301,61 @@ function MLJBase.fit(model::LDA, ::Int, X, y)
...
+    nclasses = length(class_list) 
+
     # compute the distances in the transformed space between pairs of rows
-    # the probability matrix is `n x nc` and normalised accross rows
-    P = pairwise(m.dist, XWt, centroids, dims=1)
-    # apply a softmax transformation
-    P .-= maximum(P, dims=2)
-    P  .= exp.(-P)
+    # The discriminant matrix `P` is of dimension `n x nc`
+    #  P[i,k] = -0.5*(xᵢ −  µₖ)ᵀΣ⁻¹(xᵢ −  µₖ) + log(priorsₖ) and Σ⁻¹ = I due to the nature of the projection_matrix
```
tlienart: there's some weird unicode here after `\mu`, if possible don't use it

src/MultivariateStats.jl
```diff
@@ -301,37 +301,61 @@ function MLJBase.fit(model::LDA, ::Int, X, y)
...
+
+    P = pairwise(SqEuclidean(), XWt, centroids, dims=1)
```
tlienart: it's not necessary to block  it to  SqEuclidean, please leave this a parameter

src/MultivariateStats.jl
```diff
@@ -301,37 +301,61 @@ function MLJBase.fit(model::LDA, ::Int, X, y)
...
+    P .*= -0.5
```
tlienart: remove this line, it's not necessary

src/MultivariateStats.jl
```diff
@@ -301,37 +301,61 @@ function MLJBase.fit(model::LDA, ::Int, X, y)
...
+    P .+= log.(priors)'
+    #display(P)
```
tlienart: please remove

src/MultivariateStats.jl
OkonSamuel: it superscript and subscrip unicode notation

OkonSamuel: It shows up well in an Editor

src/MultivariateStats.jl
OkonSamuel: it is following the formula in the comment in the code

src/MultivariateStats.jl
OkonSamuel: same here

src/MultivariateStats.jl
OkonSamuel: okay

src/MultivariateStats.jl
tlienart: ok, that's fine ignore my  comment

src/MultivariateStats.jl
tlienart: (I meant  the comment)

src/MultivariateStats.jl
tlienart: ok it's fine I guess let's leave it as you put it; my reasoning is as follows: this is tied with the SqEuclidean which is tied with a specific distance; I don't really like this in that I still  think it should be possible to use another distance in there; but  that distance would need to be weighed by some other scale; 

Eg for instance if you use cosine or absolute value distance; 

On the other hand I get what you're trying  to do with Bayes rule. 

so we could either

1. offer a standard bayesian path; which would force the SqEuclidean and allow all that you were suggesting
2. offer a non-standard  path where people can  specify their own distance and potentially a weighing parameter for their distance

what do  you think?

src/MultivariateStats.jl
OkonSamuel: lol. Okay

src/MultivariateStats.jl
tlienart: I guess there could be `LDA` (basically what's already there) and `BayesianLDA` (something similar but that implements things as you've done them, fixing the metric and using a gaussian model)

src/MultivariateStats.jl
tlienart: (as an aside, I don't think you need the  transpose; the broadcast is clever, please check)

src/MultivariateStats.jl
OkonSamuel: But there are times the user is lazy and wants these priors automatically estimated. Should i still remove this from feature code?

OkonSamuel:
Okay

tlienart:
tks, I'll check the branch out, will fix the conflicts and re-open a PR, will do this soon, cheers

tlienart:
ok there seems to be some confusion as I don't see the `BayesianLDA` model; maybe you were waiting for me to fix the conflict, sorry.

Ways out: start on a new branch off dev (probably the easiest) or

1. please checkout dev and pull
2. copy the `registry` folder and paste it somewhere else for now
3. checkout your branch
4. remove the `registry` folder and replace it with the one saved at step (2)

Then you should be good to go. As far as I understand the next step is:

- leave LDA as it was before (without Bayesian props)
- add a separate BayesianLDA along the lines that you suggested here

cheers

PR #119, opened 2019-11-01
Adding LDA and  QDA from sklearn + registry  update

tlienart:
see https://github.com/alan-turing-institute/MLJModels.jl/issues/57#issuecomment-548489874


tlienart:
minor fix needed to `import Distributions` instead of  `using Distributions`, will do this asap

Issue #124, opened 2019-11-07
Implement sample weight support for NearestNeighbor models. 


ablaom:
I am developing support for sample weights in MLJ but need models for testing my API. A straightforward extension for weights is possible for the NearestNeighor models. 

To implement the new API (already there in MLJBase) you just need to implement 

`fit(model, verbosity::Int, X, y, w)` in addition `fit(model, verbosity::Int, X, y)`. Or, you can just implement `fit(model, verbosity::Int, args...)` and in case `length(args)==3`, you set `w=args[3]` and use that, etc.

And you need to declare `MLJBase.support_weights(::Type{<:ModelType}) = true`.

That's it.

You can look at [here](https://github.com/alan-turing-institute/MLJModels.jl/blob/weights/src/builtins/Constant.jl) for the only existing example implemented so far, ConstantClassfier (too trivial for comprehensive testing).

In KNN there is already an instructional hyperparameter called `weights` which supports two values `:uniform` and `:distance`.  These could be retained and the weights supplied by user at fit-time are simply "super-imposed" on these. As usual, the main work here is a bit of testing to make sure things work. 

BTW: there are scikit-learn models that support sample weights, but it is not clesar that they are accessible through the julia wrapper ScikitLearn: https://discourse.julialang.org/t/does-scikitlearn-support-sample-weights-for-scikit-learn-models-that-do/30815 .

PR #125, opened 2019-11-07
Adding fit with weights

tlienart:
See #124 for context.

Not ready for merge yet as I'm still testing this.

* [x] check it doesn't break previous tests
* [x] add elementary test that it works with weights

This seems to "work"; however it'd be good to compare with sklearn or something similar to verify that the predictions are correct.

### DIFF
--- a/src/NearestNeighbors.jl
+++ b/src/NearestNeighbors.jl
@@ -70,7 +70,10 @@ end
 
 const KNN = Union{KNNRegressor, KNNClassifier}
 
-function MLJBase.fit(m::KNN, verbosity::Int, X, y)
+function MLJBase.fit(m::KNN, verbosity::Int, X, y, w=nothing)
+    if w !== nothing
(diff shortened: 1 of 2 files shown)


src/NearestNeighbors.jl
```diff
@@ -73,6 +73,7 @@ const KNN = Union{KNNRegressor, KNNClassifier}
 function MLJBase.fit(m::KNN, verbosity::Int, X, y, w=nothing)
     if w !== nothing
         length(w) == length(y) || throw(ArgumentError("The weights vector `w` must have the same length as that of the response `y`."))
+        all(w .≥ 0) || throw(ArgumentError("Only non-negative weights are supported."))
     end
```
ablaom: Dimension matching will be checked by machine constructor, so can probably drop this.
The positivity check is not there, but I will add, thanks.

ablaom:
What do you think of pulling this and adding comprehensive tests later?

tlienart:
sounds good, I'll just open an issue for the testing viz sklearn (also that would be a good issue for newcomers)

PR #128, opened 2019-11-12
+ KNNClassfier test: weights skew predictions as expected

ablaom:
This is just slightly less dumb check than the existing test to show that sample weights skew predictions in "the right direction".


tlienart:
That's great! I'm glad this works 😅

PR #137, opened 2019-11-19
Added Bayesian quadratic discriminant model

OkonSamuel:


tlienart:
Hello Sam and thanks for this again!

One thing though is that we'd rather have **fewer** models in MLJModels rather than more if that makes sense bc everything that's here needs to be maintained here.
Exposing things in MultivariateStats makes sense but here  you're extending MultivariateStats within MLJModels which is not very maintainable in the long term.

Given your recent input, would you be interested in doing a DiscriminantAnalysis.jl package, maybe a fork of the existing one  or just your own one with a different name say BayesianDiscriminantAnalysis or whatever and just have an interface to MLJ?

That way you can own that package, it's easier to maintain in the long run, and it also makes *DA models more discoverable; further it also allows to ditch the naming "Bayesian*DA*" for just "*DA" which would not be ambiguous anymore

What do you think? and I'm happy  to  help  you towards this. 

Basically what  I'm suggesting would be:

* have a dedicated and efficient package just for DA models
* profit from that  opportunity to make everything work with `n x p` convention
* make it possible to add QDA models  and other extensions that could exist
* remove the ambiguity "Bayesian*DA" and just "*DA"

OkonSamuel:
Okay.

tlienart:
Thanks! I'll close this PR then, please ping me from your package if I can help

Issue #154, opened 2019-12-04
Update RandomForestRegressor input_scitype


ablaom:
It seems this model can handle `Count` (ie integer) data just fine (in addition to `Continuous`). Probably only the order matters here, since this is a tree based method.

PR #155, opened 2019-12-05
Small fixes [failing due to LIBSVM compat bound]

tlienart:
* addresses #154 (randomforest* input scitype) + a fix for LinearRegressor in GLM

### DIFF
--- a/src/GLM.jl
+++ b/src/GLM.jl
@@ -187,7 +187,7 @@ metadata_pkg.((LinearRegressor, LinearBinaryClassifier, LinearCountRegressor),
 
 metadata_model(LinearRegressor,
     input=MLJBase.Table(MLJBase.Continuous),
-    output=AbstractVector{MLJBase.Continuous},
+    target=AbstractVector{MLJBase.Continuous},
     weights=false,
     descr=LR_DESCR
     )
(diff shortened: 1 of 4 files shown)


tlienart:
Not passing due to shitty compat bounds on LIBSVM...

src/GLM.jl
```diff
@@ -187,7 +187,7 @@ metadata_pkg.((LinearRegressor, LinearBinaryClassifier, LinearCountRegressor),
 
 metadata_model(LinearRegressor,
     input=MLJBase.Table(MLJBase.Continuous),
-    output=AbstractVector{MLJBase.Continuous},
+    target=AbstractVector{MLJBase.Continuous},
     weights=false,
```
ablaom: Good catch!

ablaom:
I don't understand this at all. 

LIBSVM.jl has no Project file, only a REQUIRE file. According to the old [julia documentation](https://docs.julialang.org/en/v0.6/manual/packages/#Requirements-Specification-1) the versions specified there are just *lower* bounds. Since it specifies "Compat 0.41.0" there should be no prohibition on Compat 3.0, right? Indeed this is confirmed by previous builds of MLJModels, where testing (of LIBSVM wrapper) carried on happily with Compat 2.2. But, if in a new environment I `add` MLJModels (which puts Compat 3.0 in the manifest) then doing `add LIBSVM` forces Compat to drop from 3.0 to 2.2. Why?

ablaom:
Just restarting the build after merging #159

PR #171, opened 2020-01-14
Ensembles for DT.jl

### LINKED ISSUE #29: Add interface for RandomForest
The current API for decisiontree.jl only deals with single trees but not with RF.

tlienart:
closes #29  (also update of the registry which I hope to have done correctly)


src/DecisionTree.jl
```diff
@@ -104,7 +109,7 @@ function get_encoding(classes_seen)
 end
 
 MLJBase.fitted_params(::DecisionTreeClassifier, fitresult) =
-    (tree_or_leaf=fitresult[1], encoding=get_encoding(fitresult[2]))
+    (tree=fitresult[1], encoding=get_encoding(fitresult[2]))
```
ablaom: Actually, DecisionTree has two distinct types, one for leaves and one for trees (which is unfortunate). If only a stump is has a different type (leaf). But I agree, the label is a little wonky.

tlienart: Yeah I thought it didn't matter... would you prefer if that was reverted?

ablaom: Nah. I've just checked the registry again for good measure. There is a problem with `Resample` which is not really a public model anyway. I've fixed it (most likely) in MLJBase and the registry update can wait till next time. 

Really good to have this new model finally added. Thanks!

Issue #181, opened 2020-01-29
Reset default regcoef parameter in LDA.


OkonSamuel:
By default the `regcoef`parameter in `LDA` and  `BayesianLDA` models from `MultivariateStats` package is set to `0`. This poses two issues
1. It throws a `PosDefException` error when fitting with the `:gsvd` method which can be confusing to users not too familiar with the model.
2. Although there are no errors when fitting with the `:whiten` method. It produces an error when the `predict` method is called.

I suggest resetting the default `regcoef` parameter to `1e-6` if using the Standard Covariance Estimator and setting to zero if using other Covariance Estimator.
@tlienart what's your opinion

tlienart:
Sounds good, are you happy to suggest a PR for this? thanks  a lot!

PR #182, opened 2020-01-29
Reset regcoef parameter

OkonSamuel:
Reset `regcoef` parameter in `LDA` and `BayesianLDA` models to a default value of `1e-6`. see: #181 
Also converted the `regcoef` type from `Real` which is used in the model struct to `Float64` used by the internal `MS.fit` method.


tlienart:
great thanks, I think it makes sense to just set the hyperparameter to  be `Float64` then instead of doing the conversion internally (it shouldn't be upon "us" to do that kind of stuff). What do you think? other than that good to  merge

OkonSamuel:
Ok. done

tlienart:
will merge, thanks!

PR #185, opened 2020-02-04
Allow Requires 1.0

DilumAluthge:
Requires 1.0 was a non breaking change, so it is fine to allow Requires 1.0.


tlienart:
Thanks!

PR #187, opened 2020-02-11
Light interface

tlienart:
This is incomplete in that I can't get the metadata tests to pass for  some reason. I.e. the first block:

```julia
@testset "metadata" begin
    @testset "reading and extracting model metadata" begin
        @test include("metadata.jl")
    end
    @testset "model search" begin
        @test include("model_search.jl")
    end
    @testset "loading model code" begin
        @test include("loading.jl")
    end
end
```

fails, primarily it fails to find "PCA" which is really odd because

```julia
julia> filter!(m -> m.name == "PCA", models())
1-element Array{NamedTuple,1}:
 (name = PCA, package_name = MultivariateStats, ... )
```

Everything else works. (failures on  CI are irrelevant, it's because some minor fixes in MLJBase haven't been synched yet)

So there's something here I don't understand but I'm not familiar enough with the registry side of thing to understand  what, I'm hoping it'd be an easy one to fix for you @ablaom 

**Next**: assuming this is easy to fix, then it can be a minor-release `0.9` which is compatible with `MLJBase 0.11`.


ablaom:
Mmm. I just re-instated the metadata tests and they pass locally. Trying again on this PR...

ablaom:
All looks good. I've also double checked with `MLJModels.check_registry()`.

So I'm merging this.

PR #200, opened 2020-03-04
Add fields in Standardizer, support callable in features field

darrencl:
`features` field is now accepting callable that returns Bool. That is, if `features(ftr)`, where `ftr` is a feature, returns true `ftr` is standardized.

Added `Bool` fields:
* `ignore` - if true, specified or evaluated features from `features` field are ignored.
* `ordered_factor` - if true, also standardizes `OrderedFactor` type
* `count` - if true, also standardizes `Count` type

### DIFF
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -322,10 +322,11 @@ MLJBase.inverse_transform(transformer::UnivariateStandardizer, fitresult, w) =
      Standardizer(; features=Symbol[], ignore=false, ordered_factor=false, count=false)
 
 Unsupervised model for standardizing (whitening) the columns of tabular data.
-If features is empty then all columns v having Continuous element scitype are standardized.
-Otherwise, the features standardized are `Continuous` named in features (`ignore=false`) or
-`Continuous` features not named in features (`ignore=true`). To allow standarization of
-`Count` or `OrderedFactor` features as well, set the appropriate flag to true.
+If features is empty then all columns `v` having Continuous element scitype are
+standardized. Otherwise, the features standardized are `Continuous` named in features
+(`ignore=false`) or `Continuous` features not named in features (`ignore=true`). To allow
+standarization of `Count` or `OrderedFactor` features as well, set the appropriate flag to
+true.
 
 Instead of supplying a features vector, a Bool-valued callable can be also be specified.
 For example, specifying `Standardizer(features = name -> name in [:x1, :x3], ignore = true,
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -319,7 +319,7 @@ MLJBase.inverse_transform(transformer::UnivariateStandardizer, fitresult, w) =
 ## STANDARDIZATION OF ORDINAL FEATURES OF TABULAR DATA
 
 """
-     Standardizer(; features=Symbol[], ignore=false)
+     Standardizer(; features=Symbol[], ignore=false, ordered_factor=false, count=false)
 
 Unsupervised model for standardizing (whitening) the columns of tabular data.
 If features is empty then all columns v having Continuous element scitype are standardized.
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -448,6 +448,8 @@ function MLJBase.fit(transformer::Standardizer, verbosity::Int, X)
     end
     fitresult_given_feature = Dict{Symbol,Tuple{Float64,Float64}}()
 
+    isempty(cols_to_fit) && @warn "No features left to standarize."
+
     # fit each feature
     verbosity < 2 || @info "Features standarized: "
     for j in cols_to_fit
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -319,7 +319,7 @@ MLJBase.inverse_transform(transformer::UnivariateStandardizer, fitresult, w) =
 ## STANDARDIZATION OF ORDINAL FEATURES OF TABULAR DATA
 
 """
-     Standardizer(; features=Symbol[], ignore=Symbol[])
+     Standardizer(; features=Symbol[], ignore=false)
 
 Unsupervised model for standardizing (whitening) the columns of
 tabular data. If `features` is empty then all columns `v` for which
@@ -343,30 +343,50 @@ otherwise, ignore specified features.
     │ 3   │ 1.14708   │ 3     │
 
 """
-@with_kw_noshow mutable struct Standardizer <: Unsupervised
-    features::Vector{Symbol} = Symbol[] # features to be standardized; empty means all
-    ignore::Vector{Symbol} = Symbol[] # features to be ignored
+mutable struct Standardizer <: Unsupervised
+    features::Vector{Symbol} # features to be standardized; empty means all
+    ignore::Bool # features to be ignored
+    ordered_factor::Bool
+    count::Bool
+end
+
+function MLJBase.clean!(transformer::Standardizer)
+    err = ""
+    if isempty(transformer.features) && transformer.ignore
+        err *= "Features to be ignored must be specified in features field."
+    end
+    return err
+end
+
+# keyword constructor
+function Standardizer(
+    ; features=Symbol[], ignore=false, ordered_factor=false, count=false)
+    transformer = Standardizer(features, ignore, ordered_factor, count)
+    message = MLJBase.clean!(transformer)
+    isempty(message) || throw(ArgumentError(message))
+    return transformer
 end
 
 function MLJBase.fit(transformer::Standardizer, verbosity::Int, X)
     all_features = Tables.schema(X).names
     mach_types   = collect(eltype(selectcols(X, c)) for c in all_features)
+    type_tup = [
+        (transformer.ordered_factor, Union{Continuous, OrderedFactor}),
+        (transformer.count, Count)
+    ]
+    additional_types = Union{last.(filter!(x->x[1], type_tup))...}
 
-    issubset(transformer.ignore, all_features) ||
-        @warn "Some ignored features not present in table to be fit. "
     # determine indices of all_features to be transformed
     if isempty(transformer.features)
         cols_to_fit = filter!(eachindex(all_features) |> collect) do j
-            !(all_features[j] in transformer.ignore) &&
-                mach_types[j] <: AbstractFloat
+            mach_types[j] <: Union{AbstractFloat, additional_types}
         end
     else
         issubset(transformer.features, all_features) ||
             @warn "Some specified features not present in table to be fit. "
         cols_to_fit = filter!(eachindex(all_features) |> collect) do j
-            all_features[j] in transformer.features &&
-                !(all_features[j] in transformer.ignore) &&
-                mach_types[j] <: Real
+            (all_features[j] in transformer.features && !transformer.ignore) &&
+                mach_types[j] <: Union{Real, additional_types}
         end
     end
 
@@ -393,6 +413,12 @@ end
 MLJBase.fitted_params(::Standardizer, fitresult) = (mean_and_std_given_feature=fitresult,)
 
 function MLJBase.transform(transformer::Standardizer, fitresult, X)
+    type_tup = [
+        (transformer.ordered_factor, Union{Continuous, OrderedFactor}),
+        (transformer.count, Count)
+    ]
+    additional_types = Union{last.(filter!(x->x[1], type_tup))...}
+
     # `fitresult` is dict of column fitresults, keyed on feature names
     features_to_be_transformed = keys(fitresult)
 
@@ -404,10 +430,13 @@ function MLJBase.transform(transformer::Standardizer, fitresult, X)
     col_transformer = UnivariateStandardizer()
 
     cols = map(all_features) do ftr
+        ftr_data = selectcols(X, ftr)
         if ftr in features_to_be_transformed
-            transform(col_transformer, fitresult[ftr], selectcols(X, ftr))
+            col_to_transform = elscitype(ftr_data) <: additional_types ?
+                coerce(ftr_data, Continuous) : ftr_data
+            transform(col_transformer, fitresult[ftr], col_to_transform)
         else
-            selectcols(X, ftr)
+            ftr_data
         end
     end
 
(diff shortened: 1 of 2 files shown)


darrencl:
More info of use case in https://github.com/alan-turing-institute/MLJ.jl/issues/455#issuecomment-593782285

src/builtins/Transformers.jl
```diff
@@ -319,13 +319,15 @@ MLJBase.inverse_transform(transformer::UnivariateStandardizer, fitresult, w) =
 ## STANDARDIZATION OF ORDINAL FEATURES OF TABULAR DATA
 
 """
-     Standardizer(; features=Symbol[])
+     Standardizer(; features=Symbol[], features_ignored=Symbol[])
```
ablaom: Finicky: I suggest `ignore` is a sufficiently informative field name.

darrencl:
@ablaom I agree, that approach is more general, but yeah, this PR is more user friendly. I think with this option the user can easily use this without reading the details in the docs, which is great. 

Hmm, I am wondering if it is a better idea to combine those, i.e. having `features` and `ignore` to accept callable. In addition to have the option of easily specifying `Vector` (like we have in this PR), the user could have more control on the features to be processed/ignored. With callable, having 2 options to be redundant, but maybe we can have something more flexible like:

- If in `features(ftr) == false`, features is ignored.
- If `ignore(ftr) == true` or `ftr in ignore` (if it's a `Vector`), features is ignored, regardless if `features(ftr) == true`

What do you reckon?

ablaom:
Actually, even the existing PR is a little flawed, now I reflect. What's the user to expect if they do `Standarizer(features=[:x], ignore=[:x])`? This should be disallowed, which means we need to define the `mutable struct` and a `clean!` method manually and not use @mlj_model. The `clean!` needs to check that `features` and `ignore` have empty intersection, no? (You can see how to do this [here](https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#New-model-type-declarations-and-optional-clean!-method-1)). 

We could allow any *one* of the fields to be callable, but not both. In the former case `clean!` can check if the elements of the non-callable all evaluate to `false` and throw an error otherwise. I don't see a way to check *two* callables for compatibility, since we don't know, a priori, the full list of feature names.

Am I making sense?

darrencl:
I initially thought that `ignore` will have higher priority than `features`, that is if `:x` is in both `features` and `ignore` it'll be ignored. 

However, yeah I think it makes more sense to ensure they don't have any intersection instead. I'll update this! :)

OkonSamuel:
@ablaom and @darrencl. Instead of having `features` and `ignore` keyword which are both `Vector{Symbol}` or callable. Don't you think its better to have `features` as a `Vector{Symbol}`and `ignore` as a `Bool` so that if `ignore` is true all features in `features` are standardized and if `ignore` is false the all features apart from those in `features` are standardized.  That way we can only have one callable.

darrencl:
@OkonSamuel @ablaom Hmm, now that I think about it, I agree with having `ignore` as `Bool` instead. If we're going to ensure that `features` and `ignore` don't have any intersect, then it's no point having both as `Vector`/callable (I think?). 

Unless we're going to allow something like:
```julia
# Standardize all integers except age
Standardizer(feature= x->typeof(x) == Int64, ignore=[:age])
```
But then again, this kind of use case requires the user to be aware on this callable argument anyway. Assuming the user to be aware of that, this can easily be replaced by 
```julia
Standardizer(feature= x->typeof(x) == Int64 && x != :age)
```

ablaom:
Thanks to both for taking the time to get this right. This is good, because this could model how we do similar things elsewhere. (cc @vollmersj FeatureSelector).

I agree the suggestion of @OkonSamuel is a good one.  Let me flush it out by writing out a proposed docstring. Note that I am also adding the possibility that `OrderedFactor`s be standardised, that is any `Ordinal=Union{Continuous,Count,OrderedFactor}` can be standardized. (This is easy to implement: `coerce(v, Continuous)` will work on any vector `v` of this `elscitype`. ) But I think the logic will be much clearer for the user (in parsing the effect of `ignore=true`) if we add new fields for this, `count=false`, `ordered_factor=false`. 

The docstring:

---

     Standardizer(; features=Symbol[], ignore=false, ordered_factor=false, count=false)

Unsupervised model for standardizing (whitening) the columns of tabular data. If `features` is empty then all columns `v` having `Continuous` element scitype are standardized. Otherwise, the features standardized  are those named in `features` (`ignore=false`) or those *not* named in `features` (`ignore=true`). To allow standarization of `Count` or `OrderedFactor` features as well, set the appropriate flag to `true`.  

Instead of supplying a `features` vector,  a `Bool`-valued callable can be also be specified. For example, specifying `Standardizer(features = name -> name in [:x1, :x3], ignore = true, count=true)` has the same effect as `Standardizer(features = [:x1, :x3], ignore = true, count=true)`, namely to standardise all `Continuous` and `Count` features, with the exception of `:x1` and `:x3`.  

---

The only thing we should probably disallow is `features=[], ingore=true`.

src/builtins/Transformers.jl
```diff
@@ -343,30 +343,50 @@ otherwise, ignore specified features.
...
+    transformer = Standardizer(features, ignore, ordered_factor, count)
+    message = MLJBase.clean!(transformer)
+    isempty(message) || throw(ArgumentError(message))
+    return transformer
 end
 
 function MLJBase.fit(transformer::Standardizer, verbosity::Int, X)
     all_features = Tables.schema(X).names
     mach_types   = collect(eltype(selectcols(X, c)) for c in all_features)
+    type_tup = [
```
ablaom: Okay, this is old code can be improved with scitype methods and schema. Let's just remove references to machine types which makes the code harder to read. Please do `scitypes = schema(X).scitypes` and remove `mach_types` altogether.

Instead of `type_up` business, can I suggest

```julia
scitypes = [Continuous]
transformer.ordered_factor && push!(scitypes, OrderedFactor)
transformer.count && push!(scitypes, Count)
AllowedScitype = Union{scitype...}
```

And then, see below...

src/builtins/Transformers.jl
```diff
@@ -343,30 +343,50 @@ otherwise, ignore specified features.
...
+    additional_types = Union{last.(filter!(x->x[1], type_tup))...}
 
-    issubset(transformer.ignore, all_features) ||
-        @warn "Some ignored features not present in table to be fit. "
     # determine indices of all_features to be transformed
     if isempty(transformer.features)
         cols_to_fit = filter!(eachindex(all_features) |> collect) do j
-            !(all_features[j] in transformer.ignore) &&
-                mach_types[j] <: AbstractFloat
+            mach_types[j] <: Union{AbstractFloat, additional_types}
```
ablaom: The condition we want here is now `scitypes[j] <: AllowedScitype`, and similarly later on.

src/builtins/Transformers.jl
darrencl: Done! Anyway, quick question - how can I generate `OrderedFactor`? I need it for the test.

Thanks! :)

ablaom: Good question:

`v = categorical(rand("kajldjf", 3), ordered=true)`  will do it. Or  `coerce(rand(Char, 3), OrderedFactor)`

test/builtins/Transformers.jl
```diff
@@ -140,9 +140,54 @@ end
...
+    Xt = MLJBase.transform(stand, f, X)
+
+    @test Xnew[1] == X[1]
+    @test Xnew[2] == X[2]
+    @test Xnew[3] == X[3]
+    @test elscitype(X[4]) == Count
+    @test MLJBase.std(Xnew[4]) ≈ 1.0
+    @test Xnew[5] == X[5]
+
+    stand = Standardizer(features= !=(:x5))
```
OkonSamuel: The function `!=(:x5)` is only available on Julia >= 1.2. Please change to `x -> !=(x, :x5)` instead

OkonSamuel:
@ablaom and @darrencl. I think we need to show an example in the doc header showing its use case.

src/builtins/Transformers.jl
```diff
@@ -341,27 +347,68 @@ names of features to be standardized.
...
+    if typeof(transformer.features) <: AbstractVector{Symbol}
+        if isempty(transformer.features)
+            cols_to_fit = filter!(eachindex(all_features) |> collect) do j
+                mach_types[j] <: AllowedScitype
+            end
+        else
+            issubset(transformer.features, all_features) ||
+                @warn "Some specified features not present in table to be fit. "
+            cols_to_fit = filter!(eachindex(all_features) |> collect) do j
+                (all_features[j] in transformer.features && !transformer.ignore) &&
```
OkonSamuel: @darrencl . I think this line always evaluates to `false` when transformer.ignore is `true`

OkonSamuel: probably an `ifelse` function might be better here

ablaom:
Maybe something along these lines (you need to insert the evaluations)

```julia
X = (ordinal1 = [1, 2, 3], 
     ordinal2 = categorical([:x, :y, :x], ordered=true),
     ordinal3 = [10.0, 20.0, 30.0],
     ordinal4 = [-20.0, -30.0, -40.0],
     nominal = categorical(["Your father", "he", "is"]))
schema(X)
stand1 = Standardizer()
transform(fit!(machine(stand1, X)), X)
stand2 = Standardizer(features=[:x3, ], ignore=true, count=true)
transform(fit!(machine(stand1, X)), X)
```

OkonSamuel:
@ablaom. I have a slight modification to The docstring:

 `Standardizer(; features=Symbol[], ignore=false, ordered_factor=false, count=false)`
Unsupervised model for standardizing (whitening) the columns of tabular data. If features is empty then all columns v having Continuous element scitype are standardized. Otherwise, the features standardized are Continuous features named in features (ignore=false) or the Continous features not named in features (ignore=true). To allow standarization of Count or OrderedFactor features as well, set the appropriate flag to true.

Instead of supplying a features vector, a Bool-valued callable can be also be specified. For example, specifying `Standardizer(features = name -> name in [:x1, :x3], ignore = true, count=true)` has the same effect as `Standardizer(features = [:x1, :x3], ignore = true, count=true)`, namely to standardise all Continuous and Count features, with the exception of `:x1` and `:x3`.


***
Based on what was coded i think this is better. Whats your opinion?

src/builtins/Transformers.jl
OkonSamuel: `ifelse(transformer.ignore, !(all_features[j] in transformer.features)  && mach_types[j] <: AllowedScitype,  (all_features[j] in transformer.features)  && mach_types[j] <: AllowedScitype )`

darrencl: You're right, I'll correct the test too. Thanks! :)

src/builtins/Transformers.jl
```diff
@@ -341,27 +347,68 @@ names of features to be standardized.
...
+                    mach_types[j] <: AllowedScitype
+            end
         end
     else
-        issubset(transformer.features, all_features) ||
-            @warn "Some specified features not present in table to be fit. "
         cols_to_fit = filter!(eachindex(all_features) |> collect) do j
-            all_features[j] in transformer.features && mach_types[j] <: Real
+            (transformer.features(all_features[j]) && !transformer.ignore) &&
```
OkonSamuel: Here too

src/builtins/Transformers.jl
```diff
@@ -319,7 +319,7 @@ MLJBase.inverse_transform(transformer::UnivariateStandardizer, fitresult, w) =
 ## STANDARDIZATION OF ORDINAL FEATURES OF TABULAR DATA
 
 """
-     Standardizer(; features=Symbol[], ignore=false)
+     Standardizer(; features=Symbol[], ignore=false, ordered_factor=false, count=false)
 
 Unsupervised model for standardizing (whitening) the columns of tabular data.
 If features is empty then all columns v having Continuous element scitype are standardized.
```
ablaom: "columns v having" -> "columns `v` having"

ablaom:
Yes to @OkonSamuel improvement to the docstring (and the minor typo)

Otherwise,  @darrencl, happy to merge when you are.

OkonSamuel:
@ablaom the julia nightly build is failing I can't seem to figure out why.

src/builtins/Transformers.jl
darrencl: Oops! Missed that, thanks! :)

darrencl:
@ablaom It's ready. Let's wait until the CI passed.

darrencl:
@ablaom @OkonSamuel Not sure why the CI in Julia 1.2 failed with on my latest commit, where I didn't touch any of the code. I also couldn't reproduce this when testing in my local machine. Any idea?
```
┌ Warning: Extremely small standard deviation encountered in standardization.
└ @ MLJModels ~/build/alan-turing-institute/MLJModels.jl/src/builtins/Transformers.jl:292
┌ Warning: Extremely small standard deviation encountered in standardization.
└ @ MLJModels ~/build/alan-turing-institute/MLJModels.jl/src/builtins/Transformers.jl:292
Standardizer: Test Failed at /home/travis/build/alan-turing-institute/MLJModels.jl/test/builtins/Transformers.jl:181
  Expression: MLJBase.std(Xnew[3]) ≈ 1.0
   Evaluated: NaN ≈ 1.0
Stacktrace:
 [1] top-level scope at /home/travis/build/alan-turing-institute/MLJModels.jl/test/builtins/Transformers.jl:181
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Test/src/Test.jl:1113
 [3] top-level scope at /home/travis/build/alan-turing-institute/MLJModels.jl/test/builtins/Transformers.jl:104
Standardizer: Test Failed at /home/travis/build/alan-turing-institute/MLJModels.jl/test/builtins/Transformers.jl:184
  Expression: MLJBase.std(Xnew[4]) ≈ 1.0
   Evaluated: NaN ≈ 1.0
Stacktrace:
 [1] top-level scope at /home/travis/build/alan-turing-institute/MLJModels.jl/test/builtins/Transformers.jl:184
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Test/src/Test.jl:1113
 [3] top-level scope at /home/travis/build/alan-turing-institute/MLJModels.jl/test/builtins/Transformers.jl:104
```

ablaom:
I think the problem is that the X[3] feature (:Neighborhood) which is ordered_factor is randomly generated, and there is a chance that it is constant, and so std=0. So add `Random.seed!(123)` at top of test module (replacing `123` by some number that ensures a non-constant X[3] if necessary). This should be there anyway for reproducibility. 

The code correctly issues a warning, so I don't think we need extra code to catch constant categorical columns. But if you feel so inclined...

ablaom:
cc @darrencl  https://github.com/JuliaRegistries/General/pull/10720

PR #201, opened 2020-03-04
XGBoost default HP num_rounds from 1 to 100

azev77:
num_rounds =100 is a much better default HP


ablaom:
My policy is generally to match default hyperparameters to what the package documentation declares them to be, because we are not in the business of documenting the third-party models we wrap. 

@azev77 Have you checked the XGBoost docs?

azev77:
1 `nrounds` is the only HP XGBoost doesn't provide a default value for. 
```R
library(xgboost)
data(agaricus.train, package='xgboost')
data(agaricus.test, package='xgboost')
train <- agaricus.train
test <- agaricus.test

b= xgboost(train$data,train$label)
Error in xgb.train(params, dtrain, nrounds, watchlist, verbose = verbose,  : 
  argument "nrounds" is missing, with no default

b=xgboost(train$data, train$label, nrounds = 100)
```

ablaom:
All good then. Thanks for the clarification and this PR!

ablaom:
Technically this is breaking, and I will hold out on merging for now. But have added it to the list #202

PR #204, opened 2020-03-08
Code cleanup

### LINKED ISSUE #203: Cleanup - Consistent code style
As per the [contributing guidelines](https://github.com/alan-turing-institute/MLJ.jl/blob/master/CONTRIBUTING.md), we should adhere to [blue-style](https://github.com/invenia/BlueStyle). 

However, some parts of the code seems to not following the guideline, e.g. 

https://github.com/alan-turing-institute/MLJModels.jl/blob/885bcc0a33a86fe2ee46340f20a9d99c351b1251/src/builtins/Transformers.jl#L252-L253

Which according to bluestyle, this should be
```
function MLJBase.inverse_transform(
    transformer::UnivariateDiscretizer, result , k::Integer
) 
```
If the contributors are expected to follow the guidelines, the code will be inconsistent. For that reason, I think we might need to do clean up at some stage to align with the style. 

Maybe the easiest is to run [JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl)? With this, the future contributors can easily run the formatter as well. Although, I think at some parts the formatter still doesn't satisfy BlueStyle? (See [this](https://github.com/domluna/JuliaFormatter.jl/issues/198) issue)

darrencl:
Ensure that the code has consistent style by using code formatter, such as JuliaFormatter.

Closes #203

### DIFF
--- a/src/builtins/Constant.jl
+++ b/src/builtins/Constant.jl
@@ -11,17 +11,17 @@ struct ConstantRegressor{D<:Distributions.Sampleable} <: Probabilistic
     distribution_type::Type{D}
 end
 
-ConstantRegressor(; distribution_type=Distributions.Normal) =
+ConstantRegressor(; distribution_type = Distributions.Normal) =
     ConstantRegressor(distribution_type)
 
-function MLJBase.fit(::ConstantRegressor{D}, verbosity::Int, X, y) where D
+function MLJBase.fit(::ConstantRegressor{D}, verbosity::Int, X, y) where {D}
     fitresult = Distributions.fit(D, y)
-    cache     = nothing
-    report    = NamedTuple()
+    cache = nothing
+    report = NamedTuple()
     return fitresult, cache, report
 end
 
-MLJBase.fitted_params(::ConstantRegressor, fitresult) = (target_distribution=fitresult,)
+MLJBase.fitted_params(::ConstantRegressor, fitresult) = (target_distribution = fitresult,)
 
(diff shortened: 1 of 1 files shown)


darrencl:
@tlienart Here's a draft PR to look at JuliaFormatter. I tested this on Constant.jl. If there's any other file you want me to test this with, kindly let me know! :)

tlienart:
did you try playing with the parameters a  bit? I like the idea that the formatter would  just get "everything in the same style" (though it  doesn't seem to like vertically  aligning  `=` signs which I find to  sometimes help with readability (e.g. in metadata); this is a matter of personal preference though, same with the way it handles brackets).

I kind of wonder whether this is worth the effort. Generally we do want something close to it but also flexibility for whatever may  (subjectively) help readability locally.

What do  you think? (and thanks!)

darrencl:
@tlienart The PR's generated from default setting, but [here](https://domluna.github.io/JuliaFormatter.jl/stable/#Formatting-Options-1)'s the list of options that we can modify.

It seems that there is no option to align `=` vertically and I agree it's nice for metadata or something like that. With that said, we generally only want this in specific block of code and we can always skip the formatter for that block (See [here](https://domluna.github.io/JuliaFormatter.jl/stable/skipping_formatting/#Skipping-Formatting-1)).

tlienart:
Oh that makes sense, well then I'm happy for you to go ahead when you have the time, and use your judgment in terms of where it makes sense to leave the formatting as is, thanks a lot!

ablaom:
I'm closing this old PR to remove clutter. Feel free to re-open.

PR #230, opened 2020-04-17
Add ContinuousEncoder

ablaom:
I propose adding a new built-in transformer, called `ContinuousEncoder`, implemented in this PR, which combines one-hot encoding with coercion of `Count` data and feature dropping to ensure a table has only `Continuous` features. Multiple issues have suggest to me that users, especially beginners, would appreciate this tool. 

**edited**

Composing this transformer with just about any supervised model delivers a model that can be applied to any data, dropping only columns that are not `Count`, `Continuous`, `OrderedFactor` or `Multiclass`. I have an idea to augment the registry with "virtual" models that are ordinary models wrapped in this transformer, to dramatically expand the number of models found by `models(matching(X, y)` - but I'll invite discussion on this proposal in a separate issue.

Below is the docstring for the new transformer proposed in this PR. I will leave this unmerged until Monday at least to allow feedback, which is very much appreciated.

A review from @OkonSamuel  or @tlienart, at their leisure, would be great. 

Oh, and this PR also add the keyword `ignore` to OneHotEncoder (logic as in `Standardizer`)

---

    ContinuousEncoder(one_hot_ordered_factors=false, drop_last=false)

Unsupervised model for arranging all features (columns) of a table to
have `Continuous` element scitype, by applying following protocol
applied to each feature `ftr`:

- If `ftr` is already `Continuous` retain it.

- If `ftr` is `Multiclass`, one-hot encode it.

- If `ftr` is `OrderedFactor`, replace it with `coerce(ftr,
Continuous)` (vector of floating point integers), unless
`ordered_factors=false` is specified, in which case one-hot
encoded it.

- If `ftr` is `Count`, replace it with `coerce(ftr, Continuous)`.

- If `ftr` is of some other element scitype, or was not observed in
  fitting the encoder, drop it from the table.

If `drop_last=true` is specified, then one-hot encoding always drops
the last class indicator column.

*Warning:* This transformer assumes that the `Multiclass` or
 `OrderedFactor` features in new data to be transformed have the same
 underlying `CategoricalVector` encodings as those during the fit.

### DIFF
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -316,25 +316,25 @@ end
 MLJBase.inverse_transform(transformer::UnivariateStandardizer, fitresult, w) =
     [inverse_transform(transformer, fitresult, y) for y in w]
 
-## STANDARDIZATION OF ORDINAL FEATURES OF TABULAR DATA
+## STANDARDIZATION OF ORDINAL FEATURES OF TABULAR DATA;;;
 
 """
     Standardizer(; features=Symbol[], ignore=false, ordered_factor=false, count=false)
 
 Unsupervised model for standardizing (whitening) the columns of
-tabular data.  If features is empty then all columns `v` having
-Continuous element scitype are standardized. Otherwise, the features
-standardized are the `Continuous` features named in features
-(`ignore=false`) or `Continuous` features not named in features
-(`ignore=true`). To allow standarization of `Count` or `OrderedFactor`
-features as well, set the appropriate flag to true.
+tabular data.  If `features` is unspecified then all columns `v`
+having Continuous element scitype are standardized. Otherwise, the
+features standardized are the `Continuous` features named in
+`features` (`ignore=false`) or `Continuous` features not named in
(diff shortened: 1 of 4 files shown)


src/builtins/Transformers.jl
```diff
@@ -626,27 +636,36 @@ end
...
 
-*Warning:* This transformer assumes that the elements of a categorical
- feature in new data to be transformed point to the same
- CategoricalPool object encountered during the fit.
+New data to be transformed may lack features present in the fit data,
+but no *new* features can be present.
+
+*Warning:* This transformer assumes that the `Multiclass` or
+ `OrderedFactor` features in new data to be transformed have the same
+ underlying `CategoricalVector` encodings as those during the fit.
```
OkonSamuel, line 661: An example here would be super helpful

ablaom: This warning is a bit out of date and I have updated it. 

I have also added examples for the OneHotEncoder and ContinuousEncoder

ablaom:
Closed in favour of #232

PR #247, opened 2020-05-02
add `models(regex::Regex)` method

aviatesk:
add `models(regex::Regex)` method, which would enable us to find models easily.

EDIT: I expands method signature to `needle::Union{AbstractString,Regex}`


src/model_search.jl
```diff
@@ -137,7 +137,17 @@ function MLJBase.models(conditions...)
     return sort!(unsorted)
 end
 
-MLJBase.models() = models(x->true)
```
aviatesk, line 140: We don't need this `method` actually, since `all(c for c in ())` will return `true` anyway.

Issue #244, opened 2020-04-29
Disintegration of MLJModels


ablaom:
Maintenance of MLJModels has become increasingly burdensome for several reasons. Perhaps the biggest problem is that it's centralised approach to providing model API implementations ("glue code") means:

- Testing takes a long time. Tolerable, just.

- [extras] must include a very large number of packages which invariably cause fatal version conflicts with the packges in [deps] during CI. (As I understand it, during a test, the [deps] are essentially pinned when [extras] are loaded.) Less tolerable.

- With the existing package manager, we have no way to specify bounds on the algorithm-providing packages (the ones in [extras]). The latest release compatible with [deps] always get's loaded. If just one (of these many) packages makes a breaking change to the "glue code", then MLJModels CI fails. 

While https://github.com/JuliaLang/Pkg.jl/issues/1285 may help with the second and third issue, I don't think that is close to being resolved. We have also observed elsewhere that code loading using Requires.jl can be slower than otherwise. (And there is https://github.com/alan-turing-institute/MLJModels.jl/issues/243)

While the plan has always been for all algorithm-providing packages to implement their MLJ interfaces *natively*, this is not going to happen quickly. In the meantime, it would be good to address the issues above. 

In discussions of the core team, @tlienart has suggested the following remedy: Move the glue code for each package X into its own repository Xglue, with its own testing, and make X an ordinary dependency of  Xglue. (The package Xglue would be purely a "utility" package and essentially invisible to general users.) 

Such migrations could be performed incrementally and I believe *each migration would trigger only a patch release*, basically because the model registry (which tracks where to find glue code) is part of MLJModels (and so is opaque to MLJ). 

I think this a good idea and propose beginning this disintegation; See [TODO list](https://github.com/alan-turing-institute/MLJModels.jl/issues/244#issuecomment-641668554).
 cc @DilumAluthge

ablaom:
See also @tlienart 's earlier post at https://github.com/alan-turing-institute/MLJ.jl/issues/276

tlienart:
Great, and just to stress it, we don't want this to be the rule, we want this to be a temporary solution and hope that the packages for which we end up writing glue code will **integrate** and **own** said glue code. 

We definitely **do not want** other package devs with brand new cool packages to suggest a glue package. They should just own the interface like ParallelKMeans, EvoTrees or MLJLinearModels

DilumAluthge:
[...]

PR #265, opened 2020-06-09
Remove ScikitLearn and update all classifiers to predict UnivariateFiniteArrays

ablaom:
In this PR:

- Dumps all ScikitLearn related code, it's interface now living at MLJScientificTypesInterface (#244)

- Registers MLJScientificTypesInterface to continue to make the models available to the MLJ user     **edit** with no change in how the user loads the models

- Updates all probabilistic classifiers to predict `UnivariateFiniteArray`s for performance improvement buy-ins (see https://github.com/alan-turing-institute/MLJBase.jl/pull/327)

The PR is breaking because the sk-learn models have a new `package_name` trait. Eg, one must now do:

```julia
@load RidgeRegressor pkg=MLJScikitLearnInterface
```


ablaom:
No, this won't do. This will be the same issue for each pkg we migrate out. MLJScikitLearnInterface should be invisible to user; the `@load` call should not change. 

I'll need to rethink this. Maybe it's as simple as getting the "glue" package name from the `load_path`, rather than the `package_name` as currently. And we revert `package_name` to "ScikitLearn".

ablaom:
Okay, this was fixable. Now the PR is non-breaking.

Issue #242, opened 2020-04-29
Improve convenience for loading models with duplicate names


ablaom:
Context: 
- [ ] https://github.com/alan-turing-institute/MLJ.jl/issues/446
- [ ] https://github.com/alan-turing-institute/MLJ.jl/issues/257

ablaom:
While we're here, check if something can be done about #243, although I suspect "disintegration" of MLJModels will be the ultimate solution.

OkonSamuel:
Something in the lines of 
```julia
@load LinearRegressor pkg=package name=ModelName
```
or 
```julia
@load LinearRegressor as ModelName pkg=package name=ModelName
```
might be helpful when the same model types are imported in a given module from different packages.

Note
In the first case the `name` keyword is optional. The user would tend to use it when he tries to import a model type that's already present and he sees the following error `ERROR: invalid redefinition of type`(This is julia's normal behaviour by the way)

ablaom:
In fact, the *function* `load` already has an option `allow_amibiguous` and the behaviour, when set to `true` is this: When a duplicate is loaded, the old struct is still accessible by the unqualified name, but the new struct must be qualified:

```julia
julia> model1 = load("RidgeRegressor", pkg="MLJLinearModels",
       allow_ambiguous = true)

julia> RidgeRegressor
RidgeRegressor

julia> model1 = load("RidgeRegressor", pkg="MultivariateStats",
       allow_ambiguous = true)
WARNING: ignoring conflicting import of MultivariateStats_.RidgeRegressor into Main

julia> info(RidgeRegressor).package_name
"MLJLinearModels"

julia> info(MLJModels.MultivariateStats_.RidgeRegressor).package_name
"MultivariateStats"
```

PR #273, opened 2020-06-29
Allow @load/load to import two models with same name from different packages

ablaom:
Context: #242

New behaviour: 

- No error thrown if you try to load a model with a name clashing with an existing one. Rather the new model type is bound to a different name which is auto-generated by default (eg, `RidgeRegressor2`) with a warning issued, or bound to a name specified by a key-word argument `name=...`. 

- Previously only `@load` returned a default instance of the model; now `load` also does this. 

New doc-strings:

---

    @load name pkg=nothing verbosity=0 name=nothing

Load the model implementation code for the model named in the first
argument into the calling module, specfying `pkg` in the case of
duplicate names. Return a model instance with default
hyper-parameters.

To bind the new model type to a name different from the first
argument, specify `name=...`.

### Examples

    @load DecisionTreeeRegressor
    @load PCA verbosity=1
    @load SVC pkg=LIBSVM name=MyOtherSVC

See also [`load`](@ref)

---

    load(name::String; pkg=nothing, modl=Main, verbosity=1, name=nothing)

Load the model implementation code for the model type with specified
`name` (first argument) into the module `modl`, specifying `pkg` if
necesssary to resolve duplicate names. Return an instance of the model type
with default hyper-parameters.

To bind the name of the model type to a name different from the first
argument, specify `name=...`.

    load(proxy; kwargs...1)

In the case that `proxy` is a return value of `traits` (ie, has the
form `(name = ..., package_name = ..., etc)`) this is equivalent to
`load(name=proxy.name, pkg=proxy.package_name; kwargs...)`.

See also [`@load`](@ref)

### Examples

    load("RidgeRegressor",
         pkg="MLJLinearModels",
         modl=MyPackage,
         name="NewRidge",
         verbosity=0)

    load(localmodels()[1])

See also [`@load`](@ref)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -45,10 +45,11 @@ Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 GLM = "38e38edf-8417-5370-95a0-9cbb8c7f171a"
 GaussianProcesses = "891a1506-143c-57d2-908e-e1f8e92e6de9"
 LIBSVM = "b1bec4e5-fd48-53fe-b0cb-9723c09d164b"
+MLJLinearModels = "6ee0df7b-362f-4a72-a706-9e79364fb692"
 NaiveBayes = "9bbee03b-0db5-5f46-924f-b5c9c21b8c60"
 NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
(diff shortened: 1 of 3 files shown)


ablaom:
This PR needs to wait for MLJBase 0.14.1 to be released.

Issue #264, opened 2020-06-09
Get rid of MLJBase from [deps]


ablaom:
Everything that supposedly depends on MLJBase should actually only depend on MLJModelInterface

ablaom:
There is an obstacle here: some MultivariateStats models have `UnivariateFinite` priors, so this is not trivial.

OkonSamuel:
@ablaom. Yeah. But MLJBase would still be needed as a test dependency

OkonSamuel:
Also `MLJBase.@load` and `MLJBase.==` stubs may have to be moved to **MLJModelInterface**

ablaom:
I think MLJBase.==, MLJBase.info are already re-exported from MJLModelInterface.

ablaom:
Then there is now ThreshholdPredictor which uses UnivariateFinite type :-(

OkonSamuel:
> Then there is now ThreshholdPredictor which uses UnivariateFinite type :-(

Yea. This issue would be resolved once https://github.com/alan-turing-institute/MLJBase.jl/issues/416 is addressed

ablaom:
Another issue is that Models uses MLJBase.info_dict in the generation of the model registry.

ablaom:
And `available_name`.

PR #276, opened 2020-06-29
update LDA models from MultivariateStats and begin process of removing MLJBase as dependency

OkonSamuel:
This PR does a couple of things.
1. Remove **LDA** models dependency on `MLJBase`  . Replaced `UnivariateFinite` with `Dict` (see #264 )
2. Fixed bug in **LDA** models. LDA fails when one of the following scenarios is true
    - the number of unique classes `nc` in training data is `1`
    -  the number of training examples `n` is less than or equals the number of unique classes `nc` found in training data.
3. Added better descriptions for **LDA** models
4. Make default model parameters for **LDA** models close with that used in `MultivariateStats` package.
5. Added more parameters to the `report` for **LDA** models


src/MultivariateStats.jl
```diff
@@ -427,9 +427,10 @@ $BayesianLDA_DESCR
                     within-class covariance estimator, is added to the diagonal
                     of Sw to improve numerical stability. This can be useful if
                     using the standard covariance estimator.
-* `priors=nothing`: For use in prediction with Bayes rule. if `priors = nothing` estimates the prior probabilities
-                    from the data else it uses a user specified Dictionary(`::AbstractDict{<:Union{Symbol, String}, <:Real}}`)
-                    of `class => probability` pairs for each unique class found in training data.
+* `priors=nothing`: For use in prediction with Baye's rule. If `priors = nothing` then `priors` are estimated
+                    from the class proportions in the training data. Otherwise it requires a vector containing 
+                    class probabilities specified by order of class levels ommitting class levels which are not
+                    present in training data.
```
ablaom, line 433: The switch to a vector sounds less safe. What's your reason for abandoning the use of a dictionary?

src/MultivariateStats.jl
```diff
@@ -317,7 +317,7 @@ For more information about the algorithm, see the paper by Li, Zhu and Ogihara,
     cov_w::CovarianceEstimator = MS.SimpleCovariance()
     cov_b::CovarianceEstimator = MS.SimpleCovariance()
     out_dim::Int     = 0::(_ ≥ 0)
-    regcoef::Float64    = 1e-6::(_ ≥ 0)
+    regcoef::Real    = 1e-6::(_ ≥ 0)
     dist::SemiMetric = SqEuclidean()
```
ablaom, line 321: Generally a concrete type is preferred for model structs. Do you have a reason for not using one here?

OkonSamuel: Ok

src/MultivariateStats.jl
OkonSamuel: I used `Vector` instead of of `Dict` for two reasons.
1. because it's much more efficient than `Dict`.
2. To avoid annotating with `Dict{Any, Float64}. (Because the class labels could be anything)
As regards the resampling issue that can be fixed.

src/MultivariateStats.jl
```diff
@@ -563,8 +595,8 @@ $BayesianSubspaceLDA_DESCR
                     to be used by `predict` and `transform` methods, automatically set if 0 is given (default).
 * `priors=nothing`: For use in prediction with Baye's rule. If `priors = nothing` then `priors` are estimated
                     from the class proportions in the training data. Otherwise it requires a vector containing 
-                    class probabilities specified by order of class levels ommitting class levels which are not
-                    present in training data.
+                    class probabilities specified in order following the ordering of levels found in the common pool
+                    of the target vector.
```
ablaom, line 600: Minor suggestion: You could be a little more explicit and say something like "... using the order given by `levels(y)`, where `y` is the target vector. "

OkonSamuel:
> Looks good. Let's hold off merging until the next minor release.

That's fine by me

src/MultivariateStats.jl
OkonSamuel: @ablaom i think this is okay.
```
* `priors=nothing`: For use in prediction with Baye's rule. If `priors = nothing` then `priors` are estimated
                    from the class proportions in the training data. Otherwise it requires a `Vector` containing 
                    class probabilities with probabilities specified using the order given by `levels(y)` where y is the
                    target vector.
```

Issue #278, opened 2020-07-15
xgboost deprecated syntax


tlienart:
```
[17:21:44] WARNING: /workspace/srcdir/xgboost/src/objective/regression_obj.cu:170: reg:linear is now deprecated in favor of reg:squarederror.
[17:21:44] WARNING: /workspace/srcdir/xgboost/src/objective/regression_obj.cu:170: reg:linear is now deprecated in favor of reg:squarederror.
[17:21:45] WARNING: /workspace/srcdir/xgboost/src/objective/regression_obj.cu:170: reg:linear is now deprecated in favor of reg:squarederror.
[17:21:45] WARNING: /workspace/srcdir/xgboost/src/objective/regression_obj.cu:170: reg:linear is now deprecated in favor of reg:squarederror.
[17:21:45] WARNING: /workspace/srcdir/xgboost/src/objective/regression_obj.cu:170: reg:linear is now deprecated in favor of reg:squarederror.
```

seems like whatever defaults we're using in the interface should be updated; could anyone have a look at the `reg` kw on the XGBoost.jl repo and what they suggest? @gd1989

PR #279, opened 2020-07-15
Update XGBoost.jl

tlienart:
ref #278 

addresses the warning message:

```
WARNING: /workspace/srcdir/xgboost/src/objective/regression_obj.cu:170: reg:linear is now deprecated in favor of reg:squarederror.
```

### DIFF
--- a/src/XGBoost.jl
+++ b/src/XGBoost.jl
@@ -155,7 +155,7 @@ function XGBoostRegressor(
     ,feature_selector="cyclic"
     ,top_k=0
     ,tweedie_variance_power=1.5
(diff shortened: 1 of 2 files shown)


ablaom:
@tlienart Thanks for investigating and for the PR1

Failing on Julia 1.0 because XGBoost now only supports julia ^1.3.

Let's resolve this for now by restricting the XGBoost test by putting it in `if VERSION ... ` block. 

@tlienart Would you be able to add that?

ablaom:
@tlienart 🙏

PR #300, opened 2020-08-26
Typofix

OkonSamuel:
fixed slight typo in `report` returned when fitting `ConstantClassifier` model


ablaom:
@OkonSamuel Good catch, thanks!

PR #315, opened 2020-10-07
Add BinaryThresholdPredictor model

OkonSamuel:
Following discussion with @vollmersj on slack, this PR adds `BinaryThresholdPredictor` model. Which is essentially a model wrapping a `Binary` `Probabilistic` classifier and predicts a `Deterministic` response from the `AbstractArray{<:UnivariateFinite, 1}` predictions made from the wrapped model using a specified `threshold`.
This much like MLR's [tuneThreshold](https://mlr.mlr-org.com/reference/tuneThreshold.html) allows users to tune probability threshold by wrapping the `BinaryThresholdPredictor` model in a `TunedModel` wrapper.

### DIFF
--- a/src/builtins/ThresholdPredictors.jl
+++ b/src/builtins/ThresholdPredictors.jl
@@ -8,11 +8,10 @@
 training the underlying `wrapped_model`. 
 `BinaryThresholdPredictor` model predictions are deterministic and depend on the specified 
 probability `threshold`(which lies in [0, 1) ). The positive class is predicted if the 
-probability value exceeds the threshold.
+probability value exceeds the threshold, where the postive class is the second class 
+returned by levels(y), if y is the target.
 
 Note:
-- The positive class is taken to be the last class in `levels(y)` where y
- is the target vector.
 - The default `threshold`(0.5) predicts the modal class.
 """
 mutable struct BinaryThresholdPredictor{M <: Probabilistic} <: Deterministic
@@ -43,6 +42,12 @@ function BinaryThresholdPredictor(;wrapped_model=ConstantClassifier(), threshold
 end
 
 function MLJBase.fit(model::BinaryThresholdPredictor, verbosity::Int, args...)
+    scitype(args[2]) <: AbstractVector{Multiclass{2}} && begin
+        first_class, second_class = levels(args[2])
+        @warn "Taking positive class as `$(second_class)` and negative class as
+        `$(first_class)`.
+        Coerce target to `OrderedFactor{2}` to suppress this warning."
+    end
     model_fitresult, model_cache, model_report = MLJBase.fit(
         model.wrapped_model, verbosity-1, args...
     )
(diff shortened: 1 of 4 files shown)


src/builtins/ThresholdPredictors.jl
```diff
@@ -0,0 +1,156 @@
...
+    )
+end
+
+function MLJBase.predict(model::BinaryThresholdPredictor, fitresult, X)
+   yhat = MLJBase.predict(model.wrapped_model, fitresult[1], X)
+   length(yhat.prob_given_ref) == 2 || begin
+   @warn "Predicted `AbstractVector{<:UnivariateFinite}`"*
+       " contains only 1 class. Hence predictions will only contain this class "*
+       "irrrespective of the set `threshold` "
+   return mode.(yhat)
```
ablaom, line 82: Nice.

test/builtins/ThresholdPredictors.jl
```diff
@@ -0,0 +1,108 @@
+module TestThresholdPredictors
+using Test, MLJModels, CategoricalArrays
+
+import MLJBase
+
+@testset "BinaryThresholdPredictor" begin
+    X = NamedTuple{(:x1,:x2,:x3)}((rand(10), rand(10), rand(10)))
+    yraw = ["MLJ", "ScikitLearn", "ScikitLearn", "ScikitLearn"]
+    yraw1 = ["ScikitLearn", "ScikitLearn", "ScikitLearn", "ScikitLearn"]
```
ablaom, line 9: lol

ablaom (commented):
@OkonSamuel Okay, this is a fantastic contribution, thanks. Feedback:

1. Since this is a wrapper, set `is_wrapper(::Type{<:BinaryThresholdPredictor}} = true` to prevent the model appearing in `models()` output. 

2. What is the reason for implementing `_predict_threshold` using `_div` and the max ratio business,  instead of ordinary `<` comparison? Is this because you have a generalization to more than two classes in mind? If so, I approve. If not, perhaps we should avoid an unnecessary division operation. 

2. I would restrict this wrapper to `OrderedFactor{2}`. The general philosophy is that there is no unambiguously defined "true" class for `Mutliclass` data.  By forcing the user to coerce to `OrderedFactor`, we have a better chance that they confront this ambiguity safely (by resetting level order appropriately). The docstring should also state explicitly what is meant by the "positive" class (see my comment above). 

An alternative would be to have `fit` throw a warning everytime it encounters `Multiclass{2}` data. Something like:

"Expected `OrderedFactor{2}` target, but got `Multiclass{2}`. Assigning \"positive\" class to the $second_class_in_levels. Coerce target to `OrderedFactor` to suppress this warning. "

src/builtins/ThresholdPredictors.jl
```diff
@@ -0,0 +1,156 @@
...
+##
+"""
+    BinaryThresholdPredictor(wrapped_model=ConstantClassifier(), threshold=0.5)
+    Wraps a `Probabilistic` model(`wrapped_model`) supporting binary classification in a 
+`BinaryThresholdPredictor` model. Training the `BinaryThresholdPredictor` model results in 
+training the underlying `wrapped_model`. 
+`BinaryThresholdPredictor` model predictions are deterministic and depend on the specified 
+probability `threshold`(which lies in [0, 1) ). The positive class is predicted if the 
+probability value exceeds the threshold.
+
```
ablaom, line 12: "where the  \"postive\" class is the *second* class returned by `levels(y)`, where `y` is the target. "

OkonSamuel:
> * What is the reason for implementing `_predict_threshold` using `_div` and the max ratio business,  instead of ordinary `<` comparison? Is this because you have a generalization to more than two classes in mind? If so, I approve. If not, perhaps we should avoid an unnecessary division operation.
> *

Yeah. I have plans to create a Multiclass version of this wrapper later on.

OkonSamuel:
> An alternative would be to have `fit` throw a warning everytime it encounters `Multiclass{2}` data. Something like:

I think i prefer this.

ablaom (approved):
Awesome.

PR #327, opened 2020-10-19
Support multivariate kNN regression

mateuszbaran:
This should add support for multivariate kNN regression. How should I modify `metadata_model` to indicate this?

### DIFF
--- a/src/NearestNeighbors.jl
+++ b/src/NearestNeighbors.jl
@@ -10,6 +10,7 @@ const MMI = MLJModelInterface
 using Distances
 
 import ..NearestNeighbors
+import ..Tables
 
 const NN = NearestNeighbors
 
@@ -141,20 +142,18 @@ function MMI.predict(m::KNNRegressor, (tree, y, w), X)
 
     for i in eachindex(idxs)
         idxs_  = idxs[i]
-        println(idxs_)
         dists_ = dists[i]
         values = [ymat[j,:] for j in idxs_]
         if w !== nothing
             w_ = w[idxs_]
         end
-        println(preds)
         if m.weights == :uniform
             preds[i,:] .= sum(values .* w_) / sum(w_)
         else
             preds[i,:] .= sum(values .* w_ .* (1.0 .- dists_ ./ sum(dists_))) / (sum(w_) - 1)
         end
     end
-    if typeof(x) <: AbstractArray
+    if typeof(y) <: AbstractArray
         return preds
     else
         return MMI.table(preds, names=Tables.schema(y).names, prototype=y)
--- a/src/NearestNeighbors.jl
+++ b/src/NearestNeighbors.jl
@@ -161,7 +161,7 @@ metadata_pkg.((KNNRegressor, KNNClassifier),
 
 metadata_model(KNNRegressor,
     input   = Table(Continuous),
-    target  = Union{AbstractVector{Continuous}, AbstractVector{<:AbstractVector{Continuous}}},
+    target  = Union{AbstractVector{Continuous}, AbstractVector{<:AbstractArray{Continuous}}},
     weights = true,
     descr   = KNNRegressorDescription
     )
--- a/src/NearestNeighbors.jl
+++ b/src/NearestNeighbors.jl
@@ -161,7 +161,7 @@ metadata_pkg.((KNNRegressor, KNNClassifier),
 
 metadata_model(KNNRegressor,
     input   = Table(Continuous),
-    target  = AbstractVector{Continuous},
+    target  = Union{Table(Continuous), AbstractVector{Continuous}},
     weights = true,
     descr   = KNNRegressorDescription
     )
--- a/src/NearestNeighbors.jl
+++ b/src/NearestNeighbors.jl
@@ -153,7 +153,7 @@ function MMI.predict(m::KNNRegressor, (tree, y, w), X)
             preds[i,:] .= sum(values .* w_ .* (1.0 .- dists_ ./ sum(dists_))) / (sum(w_) - 1)
         end
     end
-    if typeof(y) <: AbstractArray
+    if typeof(y) <: AbstractVector
         return preds
     else
         return MMI.table(preds, names=Tables.schema(y).names, prototype=y)
@@ -174,7 +174,7 @@ metadata_pkg.((KNNRegressor, KNNClassifier),
 
 metadata_model(KNNRegressor,
     input   = Table(Continuous),
-    target  = Union{AbstractVector{Continuous}, Table{Continuous}},
+    target  = Union{AbstractVector{Continuous}, Table(Continuous)},
     weights = true,
     descr   = KNNRegressorDescription
     )
--- a/src/NearestNeighbors.jl
+++ b/src/NearestNeighbors.jl
@@ -128,7 +128,7 @@ end
 function MMI.predict(m::KNNRegressor, (tree, y, w), X)
     Xmatrix     = MMI.matrix(X, transpose=true) # NOTE: copies the data
     idxs, dists = NN.knn(tree, Xmatrix, m.K)
-    preds       = zeros(length(idxs))
+    preds       = similar(y, length(idxs))
 
     w_ = ones(m.K)
 
--- a/src/NearestNeighbors.jl
+++ b/src/NearestNeighbors.jl
@@ -161,7 +161,7 @@ metadata_pkg.((KNNRegressor, KNNClassifier),
 
 metadata_model(KNNRegressor,
     input   = Table(Continuous),
-    target  = Union{Table(Continuous), AbstractVector{Continuous}},
+    target  = Union{AbstractVector{Continuous}, AbstractVector{<:AbstractVector{Continuous}}},
     weights = true,
     descr   = KNNRegressorDescription
     )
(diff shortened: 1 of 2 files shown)


mateuszbaran:
OK, now I think the target of the kNN regressor is right. Thanks @ablaom and @tlienart for help on Slack :slightly_smiling_face: . Should I change something else?

src/NearestNeighbors.jl
```diff
@@ -128,7 +128,7 @@ end
 function MMI.predict(m::KNNRegressor, (tree, y, w), X)
     Xmatrix     = MMI.matrix(X, transpose=true) # NOTE: copies the data
     idxs, dists = NN.knn(tree, Xmatrix, m.K)
-    preds       = zeros(length(idxs))
+    preds       = similar(y, length(idxs))
```
ablaom, line 132: You need to do something different here, depending on whether `y` is a vector (single target) or a *table* (multi-target). The multi-target case would be simple if it weren't for the weights. One way to proceed  is as follows: (i) convert the table `y` to a matrix, using `mat = MMI.matrix(y)` , (ii) carry on more-or-less as before to get `pred` (a matrix), and (iii) return `MMI.table(pred, names=names, prototype=y)` where `names = Tables.schema(y).names`, to manifest the result as a table with the same type as `y`. See the docstrings for `MMI.matrix` and `MMI.table` for details. Here `MMI = MLJModelInterface`. 

If you are already familiar with the Tables.jl API you might also consider a strategy using the iterator `Tables.rows(y)` that is more performant when `y` is a row-based table, but I'm not sure it be any better than my suggestion for other types (and in particular the case `y` is a `Tables.MatrixTable`. The common use cases for table input would be `DataFrame` and wrapped matrix.

test/NearestNeighbors.jl
```diff
@@ -142,7 +142,7 @@ infos[:docstring]
 infos = info_dict(knnr)
 
 @test infos[:input_scitype] == Table(Continuous)
-@test infos[:target_scitype] == Union{Table(Continuous), AbstractVector{Continuous}}
+@test infos[:target_scitype] == Union{AbstractVector{Continuous}, AbstractVector{<:AbstractArray{Continuous}}}
 infos[:docstring]
```
ablaom, line 147: The old way was correct.

ablaom (commented):
I guess I was not sufficiently clear. Either:

- `y` is a `AbstractVector{<:AbstractFloat}` (guaranteed by `scitype(y) <: AbstractVector{Continuous}`) -  the status quo

- `y` of the above form **or** `y` is any table satisfying the Tables.jl interface with columns of `AbstractFloat` eltype (guaranteed by `scitype(y) <: Table(Continuous)`) - the multi-target case

src/NearestNeighbors.jl
```diff
@@ -161,7 +161,7 @@ metadata_pkg.((KNNRegressor, KNNClassifier),
 
 metadata_model(KNNRegressor,
     input   = Table(Continuous),
-    target  = Union{AbstractVector{Continuous}, AbstractVector{<:AbstractVector{Continuous}}},
+    target  = Union{AbstractVector{Continuous}, AbstractVector{<:AbstractArray{Continuous}}},
```
ablaom, line 164: Need `Table` here as in https://github.com/alan-turing-institute/MLJMultivariateStatsInterface.jl/blob/0b996da48eeb111c95dcc7e1a375a646c4e1d900/src/models/linear_models.jl#L158

mateuszbaran:
So it is generally discouraged in MLJ to just store arrays in a column of a table, yes? I understand that I should put tables in that column.

mateuszbaran:
For example if I have a dataframe of the structure like `df = DataFrame(a = randn(10), b = randn(10), c = collect(collect.(zip(randn(10), randn(10)))))` then each element of column `c` should be converted to what exactly?

mateuszbaran:
The kNN regressor as implemented in this PR works for such dataframe.

mateuszbaran:
It looks like I should split `c` into two columns and then merge them in fit and predict.

mateuszbaran:
Some relevant documentation I've found: https://alan-turing-institute.github.io/MLJ.jl/stable/adding_models_for_general_use/#Multivariate-targets-1

ablaom:
Yes, I wrote that. Sorry, I should have pointed it out to you earlier. 

Sorry, I don't really understand your query. Is this a question about the API or how to implement it? Re the API, the main point is that if the user is going to supply a multi-target `y`, then `y` will be any table satisfying the Tables.jl interface (with all column eltypes `<:AbstractFloat`) and that is all you know. Each row of the table corresponds to a single observation and each column a different target being predicted. Your job to ensure `predict` returns a table of the same type, whatever that is, given new input features `Xnew`. I've made a suggestion about how to do this, but I don't have a strong opinion how you do it.

ablaom:
> So it is generally discouraged in MLJ to just store arrays in a column of a table, yes?

Yes, in all models currently implemented, each column of a Tables.jl table is expected to correspond to a vector of data.

ablaom:
> The kNN regressor as implemented in this PR works for such dataframe.
?

```julia
using MLJBase, MLJModels, DataFrames

X = (x1 = rand(10), x2 = rand(10))
y_columntable = (a = randn(10),
                 b = randn(10),
                 c = collect(collect.(zip(randn(10), randn(10))))) # a table
y = DataFrame(y_columntable)
model = @load KNNRegressor

fitresult, cache, report = MLJBase.fit(model, 1, X, y)
predict(model, fitresult, X)
julia> predict(model, fitresult, X)
ERROR: BoundsError: attempt to access 3-element Array{AbstractArray{T,1} where T,1} at index [[2, 3, 1, 7, 4]]
Stacktrace:
 [1] throw_boundserror(::Array{AbstractArray{T,1} where T,1}, ::Tuple{Array{Int64,1}}) at ./abstractarray.jl:541
 [2] checkbounds at ./abstractarray.jl:506 [inlined]
 [3] _getindex at ./multidimensional.jl:742 [inlined]
 [4] getindex at ./abstractarray.jl:1060 [inlined]
 [5] manipulate(::DataFrame, ::Array{Int64,1}; copycols::Bool, keeprows::Bool) at /Users/anthony/.julia/packages/DataFrames/GtZ1l/src/abstractdataframe/selection.jl:542
 [6] #select#296 at /Users/anthony/.julia/packages/DataFrames/GtZ1l/src/abstractdataframe/selection.jl:493 [inlined]
 [7] getindex at /Users/anthony/.julia/packages/DataFrames/GtZ1l/src/dataframe/dataframe.jl:468 [inlined]
 [8] getindex(::DataFrame, ::Array{Int64,1}) at ./deprecated.jl:72
 [9] predict(::MLJModels.NearestNeighbors_.KNNRegressor, ::Tuple{KDTree{StaticArrays.SArray{Tuple{2},Float64,1,2},Euclidean,Float64},DataFrame,Nothing}, ::NamedTuple{(:x1, :x2),Tuple{Array{Float64,1},Array{Float64,1}}}) at /Users/anthony/Dropbox/Julia7/MLJ/MLJModels/src/NearestNeighbors.jl:138
 [10] top-level scope at REPL[41]:1
```

Doesn't work on `y_columntable` either, which is also a Tables.jl table, because `similar(y)` throws error. 

```

**Comment by mateuszbaran** [comment]
OK, I think I see now. I will fix this PR to make it work similarly to MultivariateStats regressors. Sorry for bothering you with this, I should've read the documentation more carefully first. The API isn't intuitive for someone who is used to working with arrays instead of DataFrames and tables. By the way, `predict` worked for me because I was using it incorrectly (giving an array as `y` instead of a table).

Thanks for your help :+1:, and once again sorry for my confusion here.

**Comment by mateuszbaran** [comment]
I think this should be correct now :slightly_smiling_face: .

File: src/NearestNeighbors.jl [hunk 4]
```diff
@@ -161,7 +175,7 @@ metadata_pkg.((KNNRegressor, KNNClassifier),
 
 metadata_model(KNNRegressor,
     input   = Table(Continuous),
-    target  = Union{AbstractVector{Continuous}, AbstractVector{<:AbstractArray{Continuous}}},
+    target  = Union{AbstractVector{Continuous}, Table{Continuous}},
     weights = true,
```
ablaom, line 178: `Table(Continuous)` instead of `Table{Continuous}` https://github.com/alan-turing-institute/ScientificTypes.jl#more-on-the-table-type

src/NearestNeighbors.jl
```diff
@@ -141,20 +142,18 @@ function MMI.predict(m::KNNRegressor, (tree, y, w), X)
...
-        println(preds)
         if m.weights == :uniform
             preds[i,:] .= sum(values .* w_) / sum(w_)
         else
             preds[i,:] .= sum(values .* w_ .* (1.0 .- dists_ ./ sum(dists_))) / (sum(w_) - 1)
         end
     end
-    if typeof(x) <: AbstractArray
+    if typeof(y) <: AbstractArray
         return preds
```
ablaom, line 157: Correct but probably less confusing (and future-proof) to use `if typeof(y) <: AbstractVector`.

ablaom (approved):
Looks good to me. 

Really appreciate the help!

mateuszbaran:
Thank you for your help as well :+1:  . I've fixed these two issues.

ablaom:
Closing in favour of #328 where target branch is dev (instead of master) and model registry is updated.

Issue #22, opened 2019-05-31
lazy activation of models not working from within packages


rssdev10:
I'm trying to make a module with MLJModel:

```julia
module Abc

import XGBoost: dump_model, save, Booster

using MLJ
using MLJBase
import MLJModels

using MLJModels.XGBoost_

function __init__()
    @info "Abc"
end

end
```

but having an error `ERROR: LoadError: UndefVarError: XGBoost_ not defined`.

Looks like there is an issue with lazy activation in 
```julia
@require XGBoost = "009559a3-9522-5dbb-924b-0b6ed2b22bb9" include("XGBoost.jl")
```

One workaround I found is 
```julia
module Abc

using XGBoost
#import XGBoost: dump_model, save, Booster

using MLJ
using MLJBase
import MLJModels

include(joinpath(MLJModels.srcdir, "XGBoost.jl"))
#using MLJModels.XGBoost_

function __init__()
    @info "Abc"
end

end
```

Also I added debug output into function `__init__` of the module MLJModels and I see that this method is called twice. I have something like: 
```
[ Info: Precompiling Abc [top-level]
[ Info: MLJModels!!!
[ Info: MLJModels!!!
[ Info: Abc
```
May be it is related to a chain of `__init__` methods.

ablaom:
I'm afraid I cannot reproduce your problem:

```julia
julia> module Abc
       
       import XGBoost: dump_model, save, Booster
       
       using MLJ
       using MLJBase
       import MLJModels
       
       using MLJModels.XGBoost_
       
       function __init__()
           @info "Abc"
       end
       
       end
[ Info: Recompiling stale cache file /Users/anthony/.julia/compiled/v1.1/XGBoost/rSeEh.ji for XGBoost [009559a3-9522-5dbb-924b-0b6ed2b22bb9]
[ Info: Abc
Main.Abc

julia> using MLJ

julia> task = load_boston()
SupervisedTask @ 5…85

julia> model = Abc.XGBoostRegressor()
MLJModels.XGBoost_.XGBoostRegressor(num_round = 1,
                                    booster = "gbtree",
                                    disable_default_eval_metric = 0,
                                    eta = 0.3,
                                    gamma = 0.0,
                                    max_depth = 6,
                                    min_child_weight = 1.0,
                                    max_delta_step = 0.0,
                                    subsample = 1.0,
                                    colsample_bytree = 1.0,
                                    colsample_bylevel = 1.0,
[...]

PR #340, opened 2020-11-09
Refactor of @load

ablaom:
Addresses https://github.com/alan-turing-institute/MLJ.jl/issues/693 . 

- The intended behaviour of `@load` is not changed by this PR with this exception: By default `verbosity=1`, so that newcomers understand what the macros is doing. I think this has been a source of confusion a number of times.

```julia
julia> @load AdaBoostStumpClassifier
[ Info: For silent loading, specify `verbosity=0`. 
import MLJDecisionTreeInterface ✔
const AdaBoostStumpClassifier = MLJDecisionTreeInterface.AdaBoostStumpClassifier ✔
AdaBoostStumpClassifier(
    n_iter = 10,
    pdf_smoothing = 0.0) @036
```

- to load code without returning an instance of the new model type, there is a new macro `@loadcode`

- Support for the `load` function is removed.  Instead, for explicit importing of model types without a macro, the user may extract the appropriate load path using the `load_path` function, a model trait now overloaded to work on strings and model registry entries:

```julia
julia> load_path("PCA")
"MLJMultivariateStatsInterface.PCA"

julia> load_path("LinearRegressor", pkg="GLM")
"MLJModels.GLM_.LinearRegressor"

julia> load_path(models()[1])
"MLJScikitLearnInterface.ARDRegressor"
```
TODO:

- [x] Test `@load` works from within modules and packages (provided interface is not Requires-loaded - a known issue, see https://github.com/alan-turing-institute/MLJModels.jl/issues/22)

### DIFF
--- a/src/MLJModels.jl
+++ b/src/MLJModels.jl
@@ -17,7 +17,7 @@ import Distributions
 export @update, check_registry
 
 # from loading.jl:
-export load, @load, info
+export load, @load, info, @loadcode
 
 # from model_search:
 export models, localmodels
--- a/src/loading.jl
+++ b/src/loading.jl
@@ -69,6 +69,13 @@ macro load(name_ex, kw_exs...)
     esc(program)
 end
 
+macro loadcode(name_ex, kw_exs...)
+    program, instance_ex = _load(__module__, name_ex, kw_exs...)
+    esc(program)
+end
+
+
+
 # builds the program to be evaluated by the @load macro:
 function _load(modl, name_ex, kw_exs...)
 
@@ -171,7 +178,8 @@ function _load(modl, name_ex, kw_exs...)
     return program, instance_ex
 end
 
-## DEPRECATED
+
+## NO LONGER SUPPORTED
 
 _deperror() = error(
     "The `load` function is no longer supported. "*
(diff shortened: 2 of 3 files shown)


ablaom:
@DilumAluthge @ExpandingMan

ExpandingMan:
I was thinking maybe instead of `@loadcode` it can be called `@import`?  It seems appropriate since the behavior is closely aligned with the functionality of the actual Julia keyword `import`, but perhaps using up the namespace that much is a bit aggressive.

DilumAluthge:
`@import` seems fine to me.

ablaom:
@import it is then

OkonSamuel:
> ~@import it is then~ It would seem this is not allowed??
> 
> ```julia
> julia> macro import() quote end end
> ERROR: syntax: invalid name "import"
> Stacktrace:
>  [1] top-level scope at none:1
> ```

@ablaom  `@loadcode` sounds a little off, what of  renaming to `@loadmodel` or `@ml_import`

ExpandingMan:
Hm, yeah it might protect the keyword import.  It was probably a bad idea to shadow it in the first place.

PR #346, opened 2020-11-29
Add the optional `install_pkgs` keyword argument to the `@load` macro

DilumAluthge:
This pull request adds the optional `install_pkgs` keyword argument to the `@load` macro.

If `install_pkgs` is `true`, then any necessary packages will automatically be installed.

The default value is `install_pkgs=false`.

Example usage:
```julia
julia> using MLJModels

julia> @load NeuralNetworkClassifier install_pkgs=true
```

## Motivation

@juliohm has brought up that installing all of the necessary packages is one of the pain points for Julia beginners when they first start out using MLJ.

Therefore, we can tell these beginner users to use `install_pkgs=true`, and then the necessary packages will automatically be installed for them.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -44,9 +44,10 @@ LIBSVM = "b1bec4e5-fd48-53fe-b0cb-9723c09d164b"
 MLJDecisionTreeInterface = "c6f25543-311c-4c74-83dc-3ea6d1015661"
 MLJMultivariateStatsInterface = "1b6a4a23-ba22-4f51-9698-8599985d3728"
 NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 ScientificTypes = "321657f4-b219-11e9-178b-2701a2544e81"
 StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["DecisionTree", "Distances", "GLM", "LIBSVM", "MLJDecisionTreeInterface", "MLJMultivariateStatsInterface", "NearestNeighbors", "ScientificTypes", "StableRNGs", "Test"]
+test = ["DecisionTree", "Distances", "GLM", "LIBSVM", "MLJDecisionTreeInterface", "MLJMultivariateStatsInterface", "NearestNeighbors", "Pkg", "ScientificTypes", "StableRNGs", "Test"]
--- a/src/loading.jl
+++ b/src/loading.jl
@@ -44,7 +44,7 @@ function load_path_and_pkg(name::String; pkg=nothing)
 end
 
 """
-    @load name pkg=nothing verbosity=0 name=nothing scope=:global
+    @load name pkg=nothing verbosity=0 name=nothing scope=:global install_pkgs=false
 
 Load the model implementation code for the model named in the first
 argument into the calling module, specfying `pkg` in the case of
@@ -89,6 +89,7 @@ function _load(modl, name_ex, kw_exs...)
     verbosity = 1
     new_name = nothing
     scope = :global
+    install_pkgs = false
 
     # parse name_ex:
     name_ = string(name_ex)
@@ -111,6 +112,8 @@ function _load(modl, name_ex, kw_exs...)
             new_name = value_ex
         elseif variable_ex == :scope
             scope = value_ex
+        elseif variable_ex == :install_pkgs
+            install_pkgs = value_ex
         else
             throw(ArgumentError(warning))
         end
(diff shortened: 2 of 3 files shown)


ablaom:
I really like this proposal. 

@DilumAluthge In case `install_packages` is false and the package is not in the env, could we catch the error and issue as suggestion to specify `install_packages=true`, with an explanation with what that will actually do?

ablaom:
Maybe for consistency with `pkg=...` option, we should call this new option `install_pkgs`??

juliohm:
Alternatively one could show a yes/no option to the user of the package is
not installed. That would help with interactive sessions.

ablaom:
@juliohm Not sure I understand the suggestion here. I'm certainly not keen on interactive behaviour by default, if that is what you are suggesting. We could instead have the flag `interactive=true`, but that  would add one more step (the user needing to answer the question "Do you want to install...?"), no?

juliohm:
I was thinking about something along the lines of `FileIO` in the beginning of its development. Whenever someone tried to load a format with `load("file.myformat")` the package would figure out if the specific package was installed and would suggest to install it otherwise with a yes/no prompt. The user would just press ENTER and keep going with the original task.

Right now the user has to interrupt his/her flow of thought, jump in the package manager, and then later come back to what he/she was originally doing. Maybe a `prompt=true` option set by default could help instead of `interactive`.

ablaom:
So to flush out @juliohm 's proposal, we have a single new flag `prompt=true` (default) with this behaviour: If `prompt=true` then:

- Ask use whether to install package if not already installed. The name of the environment should probably be mentioned in the prompt
- Ask user which package to use if the same model name is duplicated in the registry 
- If either of these two scenarios occur, issue a notice that user can set `prompt=false` to overide interactive behaviour (unless `verbosity <1`, in which case no notice is given)

If `prompt=false` then keep current "throw informative error" behaviour.

@juliohm Is this close enough to what you have in mind?

This would be more convenient and beginner-friendly than the status quo. Here are the down-sides as I see them:

- Interactive behaviour by default (might cause issues when used programatically, eg in a package module)
- Mixes package management with package loading. Some might argue package management should be stay user's responsibility (dangers of "secret knowledge", etc). 

This small bit of the interface has been a frequent discussion point and changes.  Be great to get more feedback before proceeding.

@ExpandingMan @DilumAluthge @tlienart @CameronBieganek Your thoughts?

DilumAluthge:
> * Interactive behaviour by default (might cause issues when used programatically, eg in a package module)

This is a deal-breaker for me.

If we put interactive behavior into the `@load` macro, it should not be by default. E.g. if we have a kwarg `interactive`, the default value should be `interactive=false`.

If we really really need interactive behavior by default, that should go into a separate macro `@load_interactive`, where the name of this new macro makes it very clear that the behavior will be interactive. So then we can put all of this interactive functionality into the new `@load_interactive` macro, and we keep `@load` noninteractive.

ExpandingMan:
I agree that it would be nice to have some kind of convenience feature for actually installing the packages, especially since the current ecosystem consists of a large number of disparate interface packages.  Expecting users to manage these entirely on their own seems excessive, regardless of whether users are "beginners".

That said, I'm not yet sure what I think would be the best approach here.  My first thought would be to have a completely separate macro for package installation.  I also agree with your concern about the behavior being interactive "by default" (although the default would be `prompt=false` right?).  On one hand, it might be ok because once you've installed a package, it'll appear in the manifest, which means that its installation would be (or could be) fully automated in the future.  On the other, I'm concerned that this getting stuck in code without people being abundantly aware that it is potentially interactive (which they surely won't be) unfortunate shenanigans would follow.

So, at first glance, having a keyword argument maybe `install` which is `false` by default, which causes it to do `Pkg.add` if it's not already in the environment seems like a nice feature.  I would suggest you think about it a while before implementing though.  If I have more thoughts about it in the future, I'll post here.

CameronBieganek:
Now that #340 has been merged, maybe the way to go is to make `@load` default to interactive (`prompt=true` or `interactive=true`). And the MLJ user manual can recommend that beginners use `@load` and that developers and more experienced users load models directly by importing from wherever `load_path(model)` says to import from.

ablaom:
Thanks for that. As @ExpandingMan suggests, let's sleep on it.

ablaom (approved):
Looks good to me.

ablaom:
Thanks all for the feedback. Although there is no consensus, the best compromise would in my view be:

- merge the current PR as is; and

- in a second PR, add a new method `@load_interactive` (or `@iload`??) for interactive loading

@DilumAluthge Are you happy for me to merge?

DilumAluthge:
> Thanks all for the feedback. Although there is no consensus, the best compromise would in my view be:
> 
>     * merge the current PR as is; and
> 
>     * in a second PR, add a new method `@load_interactive` (or `@iload`??) for interactive loading
> 
> 
> @DilumAluthge Are you happy for me to merge?

Sounds good to me!

Issue #353, opened 2021-01-22
@load falsely reporting "Model code ... already loaded"


ablaom:
```
julia> using MLJModels

julia> @load RidgeRegressor pkg=ScikitLearn
[ Info: For silent loading, specify `verbosity=0`. 
import MLJScikitLearnInterface ✔
const RidgeRegressor = MLJScikitLearnInterface.RidgeRegressor ✔
RidgeRegressor() ✔
RidgeRegressor(
    alpha = 1.0,
    fit_intercept = true,
    normalize = false,
    copy_X = true,
    max_iter = 1000,
    tol = 0.0001,
    solver = "auto",
    random_state = nothing) @760

julia> @load RidgeRegressor pkg=MLJLinearModels
[ Info: For silent loading, specify `verbosity=0`. 
[ Info: Model code for RidgeRegressor already loaded       <----------- NOT TRUE
┌ Warning: New model type being bound to `RidgeRegressor2` to avoid conflict with an existing name. 
└ @ MLJModels ~/.julia/packages/MLJModels/vpXEM/src/loading.jl:164
import MLJLinearModels[ Info: Precompiling MLJLinearModels [6ee0df7b-362f-4a72-a706-9e79364fb692]
 ✔
const RidgeRegressor2 = MLJLinearModels.RidgeRegressor ✔
RidgeRegressor2() ✔
RidgeRegressor(
    lambda = 1.0,
    fit_intercept = true,
    penalize_intercept = false,
    solver = nothing) @358
```

Versions:

```
julia> versioninfo()
Julia Version 1.5.1
Commit 697e782ab8 (2020-08-25 20:08 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.5.0)
  CPU: Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
Environment:
  JULIA_LTS_PATH = /Applications/Julia-1.0.app/Contents/Resources/julia/bin/julia
  JULIA_PATH = /Applications/Julia-1.5.app/Contents/Resources/julia/bin/julia
  JULIA_NUM_THREADS = 5

(jl_tYCVdf) pkg> st
Status `/private/var/folders/4n/gvbmlhdc8xj973001s6vdyw00000gq/T/jl_tYCVdf/Project.toml`
[...]

PR #354, opened 2021-01-22
Add interactive version of `@load`

### LINKED ISSUE #348: Add an interactive version of @load
See the discussion at #346

ablaom:
Here called `@iload`. This resolves #348.

Also resolves minor bug #353.

User is called to interact whenever:

- The model specified appears in multiple packages (and user has not specified pkg=...); user chooses which package is desired
- The model name is already in scope (generally because a model of the same name has been loaded from a different package); user chooses to go with auto-generated name or is invited to choose a new one (checked for validity)
- The package providing the interface for the model is not in the user's environment; the user is asked to confirm installation or abort (installation is automatic if `install=true` is specified)


ablaom:
### Review notes

I'm not sure how to add testing for a user-interactive function and have not included testing in this PR.

It would therefore be good to get a more thorough review for this, if possible. To this end, in testing interactively, it is useful to note that three packages all provide a model type called `KMeans`.

So one might try `@iload KMeans` and respond in different ways, starting perhaps  with a environment with just `MLJModels` in it.

ablaom:
@DilumAluthge As you are already familiar with this code, be great if you can take a look.

cc @ExpandingMan

DilumAluthge:
At what time are you calling the `request` function? It looks to me like you are calling it at macro expansion time, which seems incorrect to me. Instead of calling the `request` function inside the body of the macro, the macro should return an expression, and the call to the `request` function should be inside that expression.

ablaom:
@DilumAluthge Many belated thanks for the review. Yes you are correct in that some of the `request` calls were not wrapped in the returned expression. 

Closing in favour of #358

Issue #306, opened 2020-09-23
Add checks in OneHotEncoding to ensure categorical pools have not changed


ablaom:
```julia
using MLJ
X = (x = coerce(["black", "white", "white", "black"], Multiclass),)

model = OneHotEncoder()
mach = machine(model, X) |> fit!

julia> transform(mach, selectrows(X, 2:3))
(x__black = [0.0, 0.0],
 x__white = [1.0, 1.0],)

julia> Xproduction = (x = coerce(["white", "white"], Multiclass),)
(x = CategoricalArrays.CategoricalValue{String,UInt32}["white", "white"],)

julia> transform(mach, Xproduction)
(x__black = [1.0, 1.0],
 x__white = [0.0, 0.0],)
```

Wrong!

ablaom:
Related: https://github.com/alan-turing-institute/MLJ.jl/issues/663

ablaom:
Resolved by #360

PR #360, opened 2021-01-29
Check onehotencoding category level

i-aki-y:
This PR tries to add a check for category level mismatch in OneHotEncoder.

As pointed out in #306, a transform function for OneHotEncoder returns wrong values when the transforming features have different category levels from fitting time.

This PR checks whether each transforming feature has the same category levels as the fitted version in the transform function and throws an exception when a mismatch is detected.

## Changes
1. Add a dictionary into the OneHotEncoderResult struct and logic to store original category levels in it.
2. Add a check whether transforming features have the same levels as the stored levels.
3. Add a test case to confirm that an exception is thrown if a mismatch is detected.
4. Add a test case for the `ignore` option of OneHotEncoder because I found it missing in the current version.

## Example
This is an example of behavior after the fix.

```julia
using MLJ

X = (x = coerce(["black", "white", "white", "black"], Multiclass),)
Xproduction = (x = coerce(["white", "white"], Multiclass),)

model = OneHotEncoder()
mach = machine(model, X) |> fit!

julia> transform(mach, selectrows(X, 2:3))
(x__black = [0.0, 0.0],
 x__white = [1.0, 1.0],)

julia> transform(mach, Xproduction)
ERROR: Found category level mismatch in feature `x`. Consider using `levels!` to ensure fitted and transforming features have the same category levels.

julia> levels!(Xproduction.x, levels(X.x))
2-element CategoricalArrays.CategoricalArray{String,1,UInt32}:
 "white"
 "white"
 
julia> transform(mach, Xproduction)
(x__black = [0.0, 0.0],
 x__white = [1.0, 1.0],)
```

### DIFF
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -1092,6 +1092,7 @@ end
 struct OneHotEncoderResult <: MMI.MLJType
     all_features::Vector{Symbol} # all feature labels
     ref_name_pairs_given_feature::Dict{Symbol,Vector{Pair{<:Unsigned,Symbol}}}
+    known_levels::Dict{Symbol, CategoricalArray}
 end
 
 # join feature and level into new label without clashing with anything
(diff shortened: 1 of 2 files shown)


i-aki-y:
It may be better just a warning instead of the error. But I'm not sure which is better. If the warning is desirable, I will fix it.

The 4th change is off-topic. If it is better to split PR, let me know.

ablaom (approved):
This is excellent, many thanks! 

Just one request on style. In MLJ, a dictionary generally has a name of the form `value_given_key`. So, for code consistency, could we change `known_levels` to `levels_given_feature` or `fitted_levels_given_feature`?

Again many thanks.

i-aki-y:
@ablaom Thank you for your review. I fix the names.

`known_levels` -> `fitted_levels_given_feature`

PR #399, opened 2021-09-03
Outlier Detection Integration

davnn:
As proposed, only removed "Abstract" from the detector types.

### DIFF
--- a/src/builtins/ThresholdPredictors.jl
+++ b/src/builtins/ThresholdPredictors.jl
@@ -4,17 +4,17 @@
 
 const THRESHOLD_SUPPORTED_ATOMS = (
     :Probabilistic,
-    :AbstractProbabilisticUnsupervisedDetector,
-    :AbstractProbabilisticSupervisedDetector)
+    :ProbabilisticUnsupervisedDetector,
+    :ProbabilisticSupervisedDetector)
 
 # Each supported atomic type gets its own wrapper:
 
 const type_given_atom = Dict(
     :Probabilistic =>
     :BinaryThresholdPredictor,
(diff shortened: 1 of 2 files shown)


ablaom:
Yeah, I'm not fond of "Abstract" in the name, but there is a reason: `ProbabilisticDetector` a  constructor for two concrete  composite types, which I've called  `ProbabilisticUnsupervisedDetector{...}` and `ProbabilisticSupervisedDetector{...}`, (see [here](https://github.com/JuliaAI/MLJOutlierDetection.jl/blob/65288eb1f25a79f4f0a454f24ba65d5fa02427f3/examples/detector_wrappers/probabilistic_detector.jl#L10)) which therefore conflict with the abstract type names you suggest. Since these are user-facing but the abstract types are not really, I went for the "Abstract" terminology. So, if instead we follow your proposal, what should we call these concrete composite types, that won't seem too weird to the user? We can't just  put "Composite" in front, as all the abstract model abstract types T get a [CompositeT version](https://github.com/JuliaAI/MLJBase.jl/blob/2f003fa9c705b88f2dd24056398d8509ad1a6a02/src/composition/abstract_types.jl#L16) in MLJBase. Maybe "ProbabilisticCompositeUnsupervisedDetector" and "ProbablisticCompositeSupervisedDetector"? Yeah, that would probably work. What do you think?

ablaom:
@davnn Thanks for this. I'll take it from here. Probably need some extra tests and I should like the author of `BinaryThresholdPredictor` to review.

ablaom:
closing in favour of #400

PR #400, opened 2021-09-07
Enhancement of BinaryThresholdPredictor to handle outlier detection models

ablaom:
This PR continues #399 and:

- Expands the types of probabilistic detectors that can be wrapped by `BinaryThresholdPredictor`, so as to include new outlier detector types added to MLJModelInterface [here](https://github.com/JuliaAI/MLJModelInterface.jl/pull/113).
- (**breaking**) Changes the field name `wrapped_model` to `model`, for consistency with the wrappers `TunedModel` and `IteratedModel`. I haven't kept backwards compatibility as a I don't think this is going to effect many users yet, but I am tagging the release as minor (breaking). @OkonSamuel may want to comment.
- Removes a check and test that are redundant. From the original test code:

```julia
function MMI.predict(model::BinaryThresholdPredictor, fitresult, X)
   yhat = MMI.predict(model.wrapped_model, fitresult[1], X)
   length(yhat.prob_given_ref) == 2 || begin
   # Due to resampling it's possible for Predicted `AbstractVector{<:UnivariateFinite}`
   # to contain one class. Hence the need for the following warning
   @warn "Predicted `AbstractVector{<:UnivariateFinite}`"*
       " contains only 1 class. Hence predictions will only contain this class "*
       "irrrespective of the set `threshold` "
 ```

Actually, it is impossible for a correctly implemented probalistic classifier to predict `UnivariateFinite` having a different number of classes from the target on which it was trained, because the pool of the `UnivariateFinite` should always coincide with the pool of the target on which the model was trained *and resampling does not effect the pool of the training target*, as in this example:

```julia
julia> y = categorical([1, 1, 1, 2]);

julia> y[1:2]
2-element CategoricalArray{Int64,1,UInt32}:
 1
 1

julia> levels(y[1:2])
2-element Vector{Int64}:
 1
 2
```
cc @davnn

@OkonSamuel Be great if you can have a look over this. Do you when you might get a chance, assuming your'e willing?


ablaom:
@davnn @OkonSamuel 

Wanting to move forward with this, I have added backwards compatibility to `BinaryThresholdPredictor` (with a depwarn for attempts to access using `wrapped_model` instead of `model`) and down-graded the  new version number to a patch.

PR #438, opened 2022-02-22
Improve inheritance for `BinaryThresholdPredictor` from atomic model

ablaom:
This PR cleans up the trait declarations for `BinaryThresholdPredictor`. In particular:

- `iteration_parameter` trait inherits from atomic model: If, say, `:n` is the atomic value, then the wrapper gets the value `:(model.n)`. Previously, this returned `:unknown` always.
- `supports_training_losses` trait now inherits from atomic model. Previously returned `false` always.
- `training_losses(model, report)` accessor function now inherits from atomic model. Previously returned `nothing` always.

These improvements help `BinaryThresholdPredictor` play better with `IteratedModel`.


OkonSamuel (approved):
LGTM

Issue #446, opened 2022-03-11
no method matching metadata_model


andreasnoack:
I'm getting the following error when trying to precompile MLJ 0.17.3 with MLJModels 0.15.5. Is there a combination of versions that is known to work?

```julia
julia> using MLJ
[ Info: Precompiling MLJ [add582a8-e3ab-11e8-2d5e-e98b27df1bc7]
ERROR: LoadError: MethodError: no method matching metadata_model(::Type{MLJModels.UnivariateFillImputer}; input_scitype=Union{AbstractVector{<:Union{Missing, ScientificTypesBase.Continuous}}, AbstractVector{<:Union{Missing, ScientificTypesBase.Count}}, AbstractVector{<:Union{Missing, ScientificTypesBase.Finite}}}, output_scitype=Union{AbstractVector{<:ScientificTypesBase.Continuous}, AbstractVector{<:ScientificTypesBase.Count}, AbstractVector{<:ScientificTypesBase.Finite}}, human_name="single variable fill imputer", load_path="MLJModels.UnivariateFillImputer")
Closest candidates are:
  metadata_model(::Any; input, target, output, weights, descr, path, input_scitype, target_scitype, output_scitype, supports_weights, docstring, load_path) at ~/.julia/packages/MLJModelInterface/GKiZ2/src/metadata_utils.jl:100 got unsupported keyword argument "human_name"
Stacktrace:
  [1] kwerr(::NamedTuple{(:input_scitype, :output_scitype, :human_name, :load_path), Tuple{Union, Union, String, String}}, ::Function, ::Type)
    @ Base ./error.jl:163
  [2] top-level scope
    @ ~/.julia/packages/MLJModels/38NmP/src/builtins/Transformers.jl:1028
  [3] include(mod::Module, _path::String)
    @ Base ./Base.jl:418
  [4] include(x::String)
    @ MLJModels ~/.julia/packages/MLJModels/38NmP/src/MLJModels.jl:1
  [5] top-level scope
    @ ~/.julia/packages/MLJModels/38NmP/src/MLJModels.jl:84
  [6] include
    @ ./Base.jl:418 [inlined]
  [7] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt64}}, source::String)
    @ Base ./loading.jl:1318
  [8] top-level scope
[...]

PR #447, opened 2022-03-11
Every version of MLJModels that has `human_name` will require at least MLJModelInterface 1.4

DilumAluthge:
See #446

The earliest version of MLJModelInterface that supports `human_name` is MLJModelInterface 1.4.0.

---

In addition to this PR, we need to go back in the registry and update the compat bounds for old versions of MLJModels.

@ablaom Can you post the list of all versions of MLJModels that use `human_name`? Then, I can put together a PR to General to update the compat bounds.


ablaom (approved):
Good catch, thanks. Can't follow my own checklist 😳

PR #449, opened 2022-03-15
Remove the section of code labelled "code to be removed in 0.15.0"

DilumAluthge:


DilumAluthge:
@ablaom A couple questions:

1. Is this PR correct? In other words, is it safe to remove these two methods?
2. If so, could you briefly explain why it is safe to remove this code?

ablaom:
1. Yes.
2. `model` used to be called `wrapped_model`. For [some time](https://github.com/JuliaAI/MLJModels.jl/blob/2ce0d7fca4a443b3d611826cce7d39a154e7e658/src/builtins/ThresholdPredictors.jl#L119) you could still call `wrapped_model` and get/set the model (and get a dep warning). It's curious that this backwards-compatibility code was already partially removed, but not fully removed. But it can all go.

PR #455, opened 2022-04-13
Add OneRuleClassifier to the registry

ablaom:
Model loads fine but we should probably wait for:

- [x] https://github.com/roland-KA/OneRule.jl/issues/4
- [x] https://github.com/roland-KA/OneRule.jl/issues/3

@roland-KA


ablaom:
Tested locally:

```julia
using MLJModels, MLJBase

M = @load OneRuleClassifier
model = M()

x1 = coerce(rand("ab", 100), Multiclass);
x2 = coerce(rand("cde", 100), Multiclass);
X = (; x1, x2);
y = x1;

mach =  machine(model, X, y) |> fit!
julia> accuracy(predict(mach, X), y)
1.0

PR #457, opened 2022-04-22
Add missing value propagation to OneHotEncoder

Chandu-4444:
The behaviour is as follows:

```julia
julia> X = (name = categorical(["a", "b", "c", "a", "b", missing]),)
(name = Union{Missing, CategoricalArrays.CategoricalValue{String, UInt32}}["a", "b", "c", "a", "b", missing],)

julia> mach = machine(OneHotEncoder(), X)
Machine{OneHotEncoder,…} trained 0 times; caches data
  model: OneHotEncoder
  args: 
    1:  Source @492 ⏎ `Table{AbstractVector{Union{Missing, Multiclass{3}}}}`


julia> fit!(mach)
[ Info: Training Machine{OneHotEncoder,…}.
[ Info: Spawning 3 sub-features to one-hot encode feature :name.
Machine{OneHotEncoder,…} trained 1 time; caches data
  model: OneHotEncoder
  args: 
    1:  Source @492 ⏎ `Table{AbstractVector{Union{Missing, Multiclass{3}}}}`


julia> transform(mach)
(name__a = Union{Missing, Float64}[1.0, 0.0, 0.0, 1.0, 0.0, missing],
 name__b = Union{Missing, Float64}[0.0, 1.0, 0.0, 0.0, 1.0, missing],
 name__c = Union{Missing, Float64}[0.0, 0.0, 1.0, 0.0, 0.0, missing],)

julia> 
```

### DIFF
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -815,7 +815,7 @@ end
 # in new input data, causing potential name clashes.
 struct OneHotEncoderResult <: MMI.MLJType
     all_features::Vector{Symbol} # all feature labels
-    ref_name_pairs_given_feature::Dict{Symbol,Vector{Pair{<:Unsigned,Symbol}}}
+    ref_name_pairs_given_feature::Dict{Symbol,Vector{Union{Pair{<:Unsigned,Symbol}, Pair{Missing, Symbol}}}}
     fitted_levels_given_feature::Dict{Symbol, CategoricalArray}
 end
 
@@ -847,9 +847,9 @@ function MMI.fit(transformer::OneHotEncoder, verbosity::Int, X)
     end
 
     ref_name_pairs_given_feature =
-        Dict{Symbol,Vector{Pair{<:Unsigned,Symbol}}}()
+        Dict{Symbol,Vector{Union{Pair{<:Unsigned,Symbol}, Pair{Missing, Symbol}}}}()
 
-    allowed_scitypes = ifelse(transformer.ordered_factor, Finite, Multiclass)
+    allowed_scitypes = ifelse(transformer.ordered_factor, Union{Missing, Finite}, Union{Missing, Multiclass})
     fitted_levels_given_feature = Dict{Symbol, CategoricalArray}()
     col_scitypes = schema(X).scitypes
     # apply on each feature
@@ -900,9 +900,7 @@ end
 
 # If v=categorical('a', 'a', 'b', 'a', 'c') and MMI.int(v[1]) = ref
 # then `_hot(v, ref) = [true, true, false, true, false]`
-_hot(v::AbstractVector{<:CategoricalValue}, ref) = map(v) do c
-    MMI.int(c) == ref
-end
+function hot(col, ref) map(col) do c  if typeof(c) != Missing MMI.int(c) == ref else missing end end end
 
 function MMI.transform(transformer::OneHotEncoder, fitresult, X)
     features = Tables.schema(X).names     # tuple not vector
@@ -928,8 +926,10 @@ function MMI.transform(transformer::OneHotEncoder, fitresult, X)
             pairs = d[ftr]
             refs = first.(pairs)
             names = last.(pairs)
-            cols_to_add = map(refs) do ref
-                float.(_hot(col, ref))
+            cols_to_add = map(refs) do ref 
+                if typeof(ref) != Missing float.(hot(col, ref)) 
+                else missing # TODO - Make a Vector that contains length(refs) * missing so that it matches the remaining encoded features.
+                end
             end
             append!(new_cols, cols_to_add)
         else
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -846,7 +846,7 @@ function MMI.fit(transformer::OneHotEncoder, verbosity::Int, X)
         end
     end
 
- 
+
     ref_name_pairs_given_feature = Dict{Symbol,Vector{Pair{<:Unsigned,Symbol}}}()
 
     allowed_scitypes = ifelse(transformer.ordered_factor, Union{Missing, Finite}, Union{Missing, Multiclass})
@@ -904,13 +904,13 @@ hot(v::AbstractVector{<:CategoricalValue}, ref) = map(v) do c
     MMI.int(c) == ref
 end
 
-function hot(col::AbstractVector{<:Union{Missing, CategoricalValue}}, ref) map(col) do c  
+function hot(col::AbstractVector{<:Union{Missing, CategoricalValue}}, ref) map(col) do c
     if ismissing(ref)
-        missing 
-    else 
-        MMI.int(c) == ref 
-    end 
-end 
+        missing
+    else
+        MMI.int(c) == ref
+    end
+end
 end
 
 function MMI.transform(transformer::OneHotEncoder, fitresult, X)
@@ -937,8 +937,8 @@ function MMI.transform(transformer::OneHotEncoder, fitresult, X)
             pairs = d[ftr]
             refs = first.(pairs)
             names = last.(pairs)
-            cols_to_add = map(refs) do ref 
-                if ismissing(ref) missing 
+            cols_to_add = map(refs) do ref
+                if ismissing(ref) missing
                 else float.(hot(col, ref))
                 end
             end
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -846,8 +846,8 @@ function MMI.fit(transformer::OneHotEncoder, verbosity::Int, X)
         end
     end
 
-    ref_name_pairs_given_feature =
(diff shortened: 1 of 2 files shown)


ablaom:
Addresses https://github.com/JuliaAI/MLJModels.jl/issues/450

src/builtins/Transformers.jl
```diff
@@ -900,9 +900,7 @@ end
 
 # If v=categorical('a', 'a', 'b', 'a', 'c') and MMI.int(v[1]) = ref
 # then `_hot(v, ref) = [true, true, false, true, false]`
-_hot(v::AbstractVector{<:CategoricalValue}, ref) = map(v) do c
-    MMI.int(c) == ref
-end
+function hot(col, ref) map(col) do c  if typeof(c) != Missing MMI.int(c) == ref else missing end end end
```
ablaom, line 903: - for readability, please unroll the `do`-`end` block, as in the original code. The same applies to `if`-`else`-`end` blocks
- Turns out that the new `hot` function is not type-stable:

```julia
 col = [1, missing, 2, 1, 2] |> categorical
julia> @code_warntype hot(col, int(col[1]))
Variables
  #self#::Core.Const(hot)
  col::CategoricalArrays.CategoricalVector{Union{Missing, Int64}, UInt32, Int64, CategoricalArrays.CategoricalValue{Int64, UInt32}, Missing}
  ref::UInt32
  #9::var"#9#10"{UInt32}

Body::Any
1 ─ %1 = Main.:(var"#9#10")::Core.Const(var"#9#10")
│   %2 = Core.typeof(ref)::Core.Const(UInt32)
│   %3 = Core.apply_type(%1, %2)::Core.Const(var"#9#10"{UInt32})
│        (#9 = %new(%3, ref))
│   %5 = #9::var"#9#10"{UInt32}
│   %6 = Main.map(%5, col)::Any                  <--------------  Any type here
└──      return %6
```

(Incidentally, it's more readable if  instead of `typeof(c) != Missing` you do `ismissing(c)` - and both compile to the same code.) Maybe to fix the instability, you want to have two `hot` methods, one dispatched on `AbstractVector{<:CategoricalVector}` and one dispatched on `AbstractVector{<:Union{Missing, CategoricalValue}}`? In each case you know at compile time the output type.

src/builtins/Transformers.jl
```diff
@@ -928,8 +926,10 @@ function MMI.transform(transformer::OneHotEncoder, fitresult, X)
             pairs = d[ftr]
             refs = first.(pairs)
             names = last.(pairs)
-            cols_to_add = map(refs) do ref
-                float.(_hot(col, ref))
+            cols_to_add = map(refs) do ref 
+                if typeof(ref) != Missing float.(hot(col, ref)) 
+                else missing # TODO - Make a Vector that contains length(refs) * missing so that it matches the remaining encoded features.
```
ablaom, line 931: if `ismissing(ref)` is easier to read - see earlier comment.

src/builtins/Transformers.jl
```diff
@@ -847,9 +847,9 @@ function MMI.fit(transformer::OneHotEncoder, verbosity::Int, X)
     end
 
     ref_name_pairs_given_feature =
-        Dict{Symbol,Vector{Pair{<:Unsigned,Symbol}}}()
+        Dict{Symbol,Vector{Union{Pair{<:Unsigned,Symbol}, Pair{Missing, Symbol}}}}()
```
ablaom, line 850: I'm a little confused by this expansion of the type. In `fit` it appears that you will never actually encounter a pair of the type `Pair{Missing, Symbol}`. See line 861. In `fit` you only really care about `levels(col)` not actual elements of `col`. Rather, it is only in `transform` that elements matter. 

Am I missing something here?

src/builtins/Transformers.jl
ablaom, line 931: Could you elaborate on the "TODO" comment. I don't quite follow. 

Are you referring to new feature to be added in a later PR?

ablaom (commented):
@Chandu-4444 This is a good start, thank you. 

Please address the comments and provide a test. The test can be pretty short but should include both options for `drop_last`. 

Let me know if you would like help with the type instability.

src/builtins/Transformers.jl
Chandu-4444: Hmm.. that's just a note I wrote while I'm working on this. I forgot to remove that. Sorry!

src/builtins/Transformers.jl
Chandu-4444: Oh yes. I almost forgot to remove this line. Turns out that I wrote this for doing [this](https://github.com/JuliaAI/MLJModels.jl/issues/450#issuecomment-1100424642).

src/builtins/Transformers.jl
Chandu-4444: > for readability, please unroll the do-end block, as in the original code. The same applies to if-else-end blocks

I've done that in the latest commit.
 
> Turns out that the new hot function is not type-stable:

I'm not completely sure of what that is. But I suppose that it has to do with the map assuming `col` as type Any. But when I checked it on my REPL I didn't find Any as the type for col I'll check it again. But I guess not it won't happen that way as I've added the types for the hot function now (as you suggested). Let me know if it's ok.

ablaom: I think this is fine now, thanks.

test/builtins/Transformers.jl
```diff
@@ -465,6 +465,57 @@ end
     @test infos[:name] == "OneHotEncoder"
     @test infos[:input_scitype] == MLJBase.Table
     @test infos[:output_scitype] == MLJBase.Table
+
+    # test the work on missing values
+    X = (name   = categorical(["Ben", "John", "Mary", "John", missing], ordered=true),
+         height = [1.85, 1.67, 1.5, 1.67],
+         favourite_number = categorical([7, 5, 10, missing, 5]),
+         age    = [23, 23, 14, 23])
+
```
ablaom, line 474: This is not a valid table because `name` has a different length from `height`. You need to pad the  outher fields with entries.

test/builtins/Transformers.jl
```diff
@@ -465,6 +465,57 @@ end
...
+    t  = OneHotEncoder()
+    f, _, report = @test_logs((:info, r"Spawning 3"),
+                         (:info, r"Spawning 3"), MLJBase.fit(t, 1, X))
+     
+    Xt = MLJBase.transform(t, f, X)
+
+    @test typeof(Xt.name__John == float.([false, true, false, true, missing])) == Missing
```
ablaom, line 481: Line 481 seems like a rather weak test. I think we should check all the elements have the expected form. How about

```julia
@test length(Xt.name__John) == 5
@test skipmissing(Xt.name__John) == float.([false, true, false, true])
@test ismissing(XT.name__John[5] )
```
?

Similarly for line 483, 505, and 507.

src/builtins/Transformers.jl
```diff
@@ -900,10 +900,19 @@ end
 
 # If v=categorical('a', 'a', 'b', 'a', 'c') and MMI.int(v[1]) = ref
 # then `_hot(v, ref) = [true, true, false, true, false]`
-_hot(v::AbstractVector{<:CategoricalValue}, ref) = map(v) do c
+hot(v::AbstractVector{<:CategoricalValue}, ref) = map(v) do c
     MMI.int(c) == ref
 end
 
+function hot(col::AbstractVector{<:Union{Missing, CategoricalValue}}, ref) map(col) do c
```
ablaom, line 907: Very minor style point: You have some extraneous whitespace at the end of line 907 and elsewhere. Your editor probably has a way to highlight superfluous whitespace and delete it. (In emacs, which I use, there is whitespace-mode).

test/builtins/Transformers.jl
Chandu-4444: Oh, this is nice. I'm very sure that I don't know many of the functions related to missing 🙂

src/builtins/Transformers.jl
Chandu-4444: Sure. I changed the settings now

test/builtins/Transformers.jl
```diff
@@ -465,6 +465,63 @@ end
...
+    f, _, report = @test_logs((:info, r"Spawning 2"),
+                        (:info, r"Spawning 2"), MLJBase.fit(t, 1, X))
+
+    Xt = MLJBase.transform(t, f, X)
+
+    @test length(Xt.name__John) == 5
+    @test collect(skipmissing(Xt.name__John)) == float.([false, true, false, true])
+    @test ismissing(Xt.name__John[5])
+    @test Xt.height == X.height
+    @test ismissing(Xt.favourite_number__5) == 5
```
ablaom, line 513: This test is causing your fail. I guess you mean
```suggestion
    @test ismissing(Xt.favourite_number__5[4])
```
?

ablaom (approved):
Looks good to me.

Thank you @Chandu-4444 for this PR. Great to have this.

PR #460, opened 2022-05-06
`matching(X, y)` should print a warning if `y` is a table and `y` only has one column

DilumAluthge:
See https://github.com/alan-turing-institute/MLJ.jl/issues/937 for motivation.

### DIFF
--- a/src/model_search.jl
+++ b/src/model_search.jl
@@ -251,10 +251,10 @@ end
 
 not_missing_and_true(x) = !ismissing(x) && x
 
-const WARN_MULTITARGET = "y is a table with only one column. " *
-                         "If y is a table, we assume you want to do multi-target modeling." *
+const WARN_MULTITARGET = "'y' is a table with only one column. " *
+                         "If 'y' is a table, we assume you want to do multi-target modeling." *
                          "If you actually want to do single-target modeling, " *
-                         "you need to convert y to a Vector."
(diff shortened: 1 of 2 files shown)


DilumAluthge:
cc: @Carlos-Trebbau

src/model_search.jl
```diff
@@ -251,11 +251,32 @@ end
...
-matching(X, y, w::AbstractDict) =
-    Checker{true,false,true,scitype(X),scitype(y)}()
+function warn_if_single_column(y)
+    if Tables.istable(y) && (length(Tables.columnnames(y)) == 1)
+        msg = string(
+            "y is a table with only one column. ",
+            "If y is a table, we assume you want to do multi-target modeling.",
+            "If you actually want to do single-target modeling, ",
+            "you need to convert y to a Vector.",
+        )
```
ablaom, line 261: Please enclose variable names in single quotes.

ablaom (changes requested):
Thanks for this, @DilumAluthge . This is a good implementation. 

Can you please make the warning string a global constant, but defined in the same file, close to where you use it. You could call it `WARN_MULTITARGET`, say. And can you please add a `log_test` that references that constant?

src/model_search.jl
DilumAluthge: Done

ablaom (approved):
Beautiful.

PR #468, opened 2022-08-03
fix one hot encoder when first value in column is missing

olivierlabayle:
Fix: https://github.com/JuliaAI/MLJModels.jl/issues/467


ablaom (approved):
Perfect, @olivierlabayle . Thanks for your identification, diagnosis and resolution. 🙏🏾

PR #473, opened 2022-09-03
make relocatable

jbrea:


ablaom (commented):
I'm not super familiar with the issue this is resolving but have no objections provided someone with expertise can approve. @DilumAluthge I'm guessing you understand this and can review please?

DilumAluthge:
I'm actually not super familiar (yet) with RelocatableFolders. @michaelhatherly Could you review this?

MichaelHatherly (approved):
Yes, looks to be doing the equivalent, in a relocatable way.

I'm not familiar with the reasoning behind loading `REGISTRY_PROJECT` at module init, I'd assume you could also implement this by just parsing that `Project.toml` during precompilation instead perhaps? (And using `include_dependency` to ensure recompilation if it changes.)

ablaom:
@MichaelHatherly Thanks for the review!

PR #478, opened 2022-09-13
add InteractionTransformer

olivierlabayle:
This PR provides an `InteractionTransformer` as suggested in https://github.com/JuliaAI/MLJModels.jl/issues/476.

Currently the model has hyperparameters types:
- order::Int
- colnames::Union{Nothing, Vector{Symbol}}

The second one may not be the most appropriate and happy to get some advice on it.

Otherwise happy to get your review comments and implement any missing interface function if any.

### DIFF
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -1022,6 +1022,50 @@ function MMI.transform(transformer::ContinuousEncoder, fitresult, X)
 end
 
 
+# # INTERACTION TRANSFORMER
+
+@with_kw_noshow mutable struct InteractionTransformer <: Static
+    order::Int                                          = 2
+    colnames::Union{Nothing, Vector{Symbol}}            = nothing
+    function InteractionTransformer(order, colnames)
+        order >= 2 || throw(ArgumentError("The order of interactions should be at least 2."))
+        if colnames !== nothing
+            length(colnames) > 1 || throw(ArgumentError("At least 2 columns are required to compute interactions."))
+        end
+        return new(order, colnames)
+    end
+end
+
+infinite_scitype(col) = eltype(scitype(col)) <: Infinite
+
+actualcolumns(colnames::Nothing, table) =
+    filter(colname -> infinite_scitype(Tables.getcolumn(table, colname)), Tables.columnnames(table))
+
+function actualcolumns(colnames::Vector{Symbol}, table)
+    diff = setdiff(colnames, Tables.columnnames(table))
+    diff != [] && throw(ArgumentError(string("Column(s) ", join([x for x in diff], ", "), " are not in the dataset.")))
+
+    for colname in colnames
+        infinite_scitype(Tables.getcolumn(table, colname)) || throw(ArgumentError("Column $colname's scitype is not Infinite."))
+    end
+    return Tuple(colnames)
+end
+
+interactions(columns, order::Int) = 
+    collect(Iterators.flatten(combinations(columns, i) for i in 2:order))
+
+interactions(columns, variables...) =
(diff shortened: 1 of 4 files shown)


src/builtins/Transformers.jl
```diff
@@ -1022,14 +1022,58 @@ function MMI.transform(transformer::ContinuousEncoder, fitresult, X)
...
+    colnames::Union{Nothing, Vector{Symbol}}            = nothing
+    function InteractionTransformer(order, colnames)
+        order >= 2 || throw(ArgumentError("The order of interactions should be at least 2."))
+        if colnames !== nothing
+            length(colnames) > 1 || throw(ArgumentError("At least 2 columns are required to compute interactions."))
+        end
+        return new(order, colnames)
+    end
+end
+
```
ablaom, line 1038: I suggest you put the hyperparameter checks in a `clean!` method and not have an internal constructor. The `clean!` method gets called at every `fit!`, so if user mutates hyperparameters illegally, this is caught. See eg, [here](https://github.com/JuliaAI/MLJTuning.jl/blob/e7e54bd14a03ff6179402f8d9eb918b018492870/src/tuned_models.jl#L360). 

Instead of `@with_kw_noshow` you can use `@mlj_model` (see [here](https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#Macro-shortcut)) and then that will interpolate a `clean!` call into the generated keyword argument constructor as well.

src/builtins/Transformers.jl
```diff
@@ -2051,3 +2100,59 @@ julia> transform(mach, x)
...
 UnivariateTimeTypeToContinuous
+
+"""
+$(MLJModelInterface.doc_header(InteractionTransformer))
+
+Generates all polynomial interaction terms up to the given order for the subset of chosen columns. 
+Any column that contains elements with scitype `<:Infinite` is a valid basis to generate interactions. 
+If `colnames` is not specified, all such columns with scitype `<:Infinite` in the table are used as a basis.
+
+
```
ablaom, line 2111: A few small details missing in the docstring, but I can add those before merge.

src/builtins/Transformers.jl
```diff
@@ -1022,14 +1022,58 @@ function MMI.transform(transformer::ContinuousEncoder, fitresult, X)
 end
 
 
+# # INTERACTION TRANSFORMER
+
+@with_kw_noshow mutable struct InteractionTransformer <: Static
+    order::Int                                          = 2
+    colnames::Union{Nothing, Vector{Symbol}}            = nothing
```
ablaom, line 1029: I think elsewhere we generally use `features` instead of `colnames`.

ablaom (commented):
Looking good.

src/builtins/Transformers.jl
olivierlabayle: Thanks for that, I have changed in favor of the `@mlj_model` macro as suggested

src/builtins/Transformers.jl
olivierlabayle: Allright, I'll let you adress that then

src/builtins/Transformers.jl
olivierlabayle: done

olivierlabayle:
> Looking good.

I think I've addressed your comments, please let me know if there is anything else!

ablaom (approved):
Good to go. Thanks @olivierlabayle for this valuable contribution.

ablaom:
Closed as merged onto dev here: https://github.com/JuliaAI/MLJModels.jl/commit/6a58242951dea0a45eebb03ba022ae0611bd2cc7

PR #519, opened 2023-06-28
Add two `SIRUS.jl` models

rikhuijzer:
It took a bit longer than expected, but finally regression was implemented for `SIRUS.jl`. I've tried to explain at the [Julia Discourse](https://discourse.julialang.org/t/ann-sirus-jl-v1-2-interpretable-machine-learning-via-rule-extraction/100932) why this is a useful model. Hopefully the `StableRulesClassifier` and `StableRulesRegressor` can be added to `MLJModels.jl` :smile:

PR #524, opened 2023-07-31
add sirus models to registry

OkonSamuel:
continuation of #519


OkonSamuel:
> @OkonSamuel why was a new PR by the way necessary?

We added a models from `SymbolicRegression.jl` after you made your initial PR. I decided to open a new PR to be safe.

PR #526, opened 2023-09-15
Fix SIRUS `load_path`s

rikhuijzer:
I've made a mistake in https://github.com/JuliaAI/MLJModels.jl/pull/524, which causes the following (https://github.com/rikhuijzer/SIRUS.jl/issues/54):

```julia
julia> RulesClassifier = @load StableRulesClassifier pkg="SIRUS" verbosity=0
SIRUS.MLJImplementation.StableForestClassifier
```
Note that `StableForestClassifier` in the output should have been `StableRulesClassifier`.

This PR fixes these mistakes. Apologies for the noise.

PR #527, opened 2023-09-20
Add a workaround to fix CI failure for Julia 1.6

ablaom:
The failure noted in #526 appears to be due to https://github.com/JuliaLang/Pkg.jl/issues/3628 . This PR adds a workaround to address the issue.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -45,7 +45,6 @@ Tables = "0.2,1.0"
 julia = "1.6"
 
 [extras]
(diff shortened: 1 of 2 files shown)


rikhuijzer:
> The failure noted in #526 appears to be due to [JuliaLang/Pkg.jl#3628](https://github.com/JuliaLang/Pkg.jl/issues/3628) . This PR adds a workaround to address the issue.

Thanks for looking into this 😄

PR #531, opened 2023-10-06
Add models from Imbalance.jl and MLJBalancing.jl and update the model registry

ablaom:
To do after this PR:

- Bump version

cc @EssamWisam


ablaom:
Looks like the `BalancedModel` got excluded automatically, which is fine. My guess is that this is because `is_wrapper` trait was set to `true`. I had forgotten about that.

PR #529, opened 2023-09-22
Remove now redundant sckitlearn hack

ablaom:

PR #540, opened 2024-01-09
Disable redundant workaround in check_registry.yml

ablaom:
Replaces #529


ablaom:
These tests won't actually suffice because `check_registry` action appears to only run on commits to master. So we'll see...

PR #550, opened 2024-03-04
Overload `save` and `restore` to address serialization issue for `BinaryThresholdPredictor`

ablaom:
This PR addresses part of https://github.com/alan-turing-institute/MLJ.jl/issues/1099

### DIFF
--- a/test/builtins/ThresholdPredictors.jl
+++ b/test/builtins/ThresholdPredictors.jl
@@ -320,7 +320,7 @@ end
     seekstart(io)
     mach2 = MLJBase.machine(io)
     close(io)
-    @test_broken MLJBase.predict(mach2, (; x = rand(2))) ==  yhat
+    @test MLJBase.predict(mach2, (; x = rand(2))) ==  yhat
 end
 
 end # module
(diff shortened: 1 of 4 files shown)


test/builtins/ThresholdPredictors.jl
```diff
@@ -268,6 +268,61 @@ MMI.input_scitype(::Type{<:NaiveClassifier}) = Table(Continuous)
...
 end
+
+
+# define a probabilistic classifier with non-persistent `fitresult`, but which addresses
+# this by overloading `save`/`restore`:
+thing = []
+struct EphemeralClassifier <: MLJBase.Probabilistic end
+function MLJBase.fit(::EphemeralClassifier, verbosity, X, y)
+    # if I serialize/deserialized `thing` then `view` below changes:
+    view = objectid(thing)
```
OkonSamuel, line 279: can we use a different variable name? Then we should add in the code comments that  `object_id` is the object's hash value. Which changes when the object is serialized.

OkonSamuel (changes requested):
Looks good to me. Just some minor points

PR #555, opened 2024-04-30
standardize AbstractFloats

tiemvanderdeure:
Changes the float type of the fitresult of `Standardizer` to `AbstractFloat`. This means standardizers will no longer force floats to `Float64`.

I also removed the second definition of `fitresult_given_feature` - I think this must have been an oversight.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,6 @@ jobs:
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
-      - uses: codecov/codecov-action@v1
+      - uses: codecov/codecov-action@v3
         with:
           file: lcov.info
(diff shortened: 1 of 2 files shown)


ablaom:
Thanks for this. It seems that codecov is hanging. I've been having this issue elsewhere. Can you try updating the codecov GH action to version @v3. You can do that here: https://github.com/JuliaAI/MLJModels.jl/blob/0de3aec7063f49a56d8f23252a42a4af896c599e/.github/workflows/ci.yml#L45

Don't use @v4 as that requires us to add a security token.

ablaom:
Could you please also fix this line:

https://github.com/JuliaAI/MLJModels.jl/blob/17578022018f08d31835a13021c52ee930605455/src/builtins/Transformers.jl#L506

More correct is `eps(T)`, right?

ablaom (approved):
Good to have this, thanks. 🙏🏾

tiemvanderdeure:
> https://github.com/JuliaAI/MLJModels.jl/blob/17578022018f08d31835a13021c52ee930605455/src/builtins/Transformers.jl#L506
> 
> More correct is `eps(T)`, right?

`eps(T)` would fail if `T` is an integer, but `eps(std(v))` should always work.

PR #566, opened 2024-07-05
add datatype to OneHotEncoder and ContinuousEncoder

tiemvanderdeure:
allows users to specify what datatype `OneHotEncoder` and `ContinuousEncoder` return, so that this now works:
```julia
using MLJModels, MLJBase
X = (x = rand(10), c = categorical(rand(Bool, 10)))
mach = machine(ContinuousEncoder(datatype = Float32), X)
fit!(mach)
transform(mach, X)
```

see discussion here: https://github.com/JuliaAI/MLJModels.jl/issues/565

There is probably a way to fix this for `ContinousEncoder` without copying data around just as much. I'll figure that out if we decide to implement this.

### DIFF
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -732,6 +732,7 @@ end
     drop_last::Bool            = false
     ordered_factor::Bool       = true
     ignore::Bool               = false
+    datatype::DataType         = Float64  
 end
 
 # we store the categorical refs for each feature to be encoded and the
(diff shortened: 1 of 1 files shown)


ablaom:
A helpful discussion, thanks. 

I now vote for adding this to MLJFlux. There really isn't another package with an MLJ interface that I can think of that is fussy about float types. If there are unforeseen cases, then you can always insert a pipeline element easily enough. For example, if you're starting with a table with all `Continuous` columns, insert `float32ify` where

```julia
float32fy(X) = Tables.tables(Float32.(Tables.matrix(X)))
```

Right? And it's overkill to create a whole `Static` model to generalise this. 

Instead of implementing with a Flux layer, I would add this to the MLJFlux data-preprocessing step (`collate` or `reformat` - I forget which this is) to avoid repeating the conversion over and over. It could be controlled with a new hyperparameter and I'd even support `Float32` as the default, even though this is technically breaking.

PR #574, opened 2024-11-24
Add affinity propagation

Yuan-Ru-Lin:
This is a companion PR of https://github.com/JuliaAI/MLJClusteringInterface.jl/pull/28.

Without this, `ap = @load AffinityPropagation pkg=Clustering` wouldn't work.


ablaom:
Oh, wow. I usually take care of this, but if it works..., thanks. 

Other steps someone needs to take  are:

- [ ] Add a [model descriptor](https://github.com/JuliaAI/MLJ.jl/blob/dev/docs/ModelDescriptors.toml) for MLJ documentation (the "Model Browser")

- [ ] Add to the [list of supported models](https://github.com/JuliaAI/MLJ.jl/blob/dev/docs/src/list_of_supported_models.md) in documenation.

Yuan-Ru-Lin:
- [x] Add a [model descriptor](https://github.com/JuliaAI/MLJ.jl/blob/dev/docs/ModelDescriptors.toml) for MLJ documentation (the "Model Browser")
- [x] Add to the [list of supported models](https://github.com/JuliaAI/MLJ.jl/blob/dev/docs/src/list_of_supported_models.md) in documenation.

Done in https://github.com/JuliaAI/MLJ.jl/pull/1147.

src/MLJModels.jl
```diff
@@ -72,6 +72,7 @@ include("metadata.jl")
 
 # read in the metadata:
 metadata_file = joinpath(srcdir, "registry", "Metadata.toml")
+Base.include_dependency(metadata_file)
```
ablaom, line 75: Not sure what the point of this may be. We've never needed it in the past. I wonder if this is responsible for the fail on julia 1.11

ablaom: @Yuan-Ru-Lin Can you please try removing this line?

Yuan-Ru-Lin: The line was for `MLJModels.jl` to be precompiled after I changed `Metadata.toml`. I recommend we keep it there because anyone who wants to add new models will benefit from it.

As for the CI error, it seems that Julia 1.11.2 expects `Statistics@1.11.2` but the latest version is `Statistics@1.11.1`.

```
...
│ To reproduce this CI run locally run the following from the same repository state on julia version 1.11.2:
│
│ `import Pkg; Pkg.test(;coverage=true, julia_args=["--check-bounds=yes", "--compiled-modules=yes"], force_latest_compatible_version=false, allow_reresolve=true)`
...
ERROR: LoadError: Unsatisfiable requirements detected for package Statistics [10745b16]:
 Statistics [10745b16] log:
 ├─possible versions are: 1.11.0 - 1.11.1 or uninstalled
 ├─restricted to versions [0.0.0, 1] by project [e5070b9b], leaving only versions: 1.11.0 - 1.11.1
 │ └─project [e5070b9b] log:
 │   ├─possible versions are: 0.0.0 or uninstalled
 │   └─project [e5070b9b] is fixed to version 0.0.0
 └─restricted to versions 1.11.2 by an explicit requirement — no versions left
```

ablaom: > The line was for MLJModels.jl to be precompiled after I changed Metadata.toml. 

Ah, I see. You're right that's a useful function doing exactly what we want- force complilation if the metadata changes. 

@DilumAluthge Do you think the fail referenced above is a Julia bug? I couldn't find any reported matching it.

DilumAluthge: @ablaom I think I've identified the problem. I'll work on a PR to fix it.

DilumAluthge: @ablaom Fixed by #576.

ablaom: Thanks @DilumAluthge. See my comment at #576.

ablaom:
@Yuan-Ru-Lin Are you happy to rebase, now I've merged #576. 

While you're at it, can you please bump Julia compat to 1.10 and bump change "1.6" in CI.yml to "1.10"?

ablaom:
And if you're happy for me to merge, please remove the "Draft" tag, thanks.

Yuan-Ru-Lin:
@ablaom all done!

DilumAluthge:
@ablaom Is it necessary to drop support for Julia 1.6?

ablaom:
> @ablaom Is it necessary to drop support for Julia 1.6?

Mmm. Not strictly necessary, but this is already happening in the MLJ ecosystem elsewhere. I think that narrow bandwidth for maintenance, it's hard to justify supporting < LTS. Do you have a specific reason for lamenting the loss of support?

PR #576, opened 2024-12-12
The issue with stdlib versions being fixed to 0.0.0 has been fixed in new versions of Julia

DilumAluthge:


test/runtests.jl
```diff
@@ -1,7 +1,12 @@
-# The next two lines added as a workaround to
-# https://github.com/JuliaLang/Pkg.jl/issues/3628 (Julia 1.6):
 import Pkg
-Pkg.add(name="Statistics", version=VERSION)
+
+if Base.VERSION >= v"1.10-"
```
ablaom, line 3: Thanks for the help here!!

Help me understand why this fixes the issue identified at #574, which is a Julia 1.11.2 failure, as far as I understand it.

ablaom: Ok. Weird. 1.11.2 tests are passing. I'm going to merge and anyway.

